Issue #2, opened 2015-10-25
jsonrpc field missing from request


ecksun:
I'm doing my first project in rust and when trying to use this library I noticed that the jsonrpc field is missing from the request, using wireshark:

```
POST /jsonrpc HTTP/1.1
Host: some.host
Content-Length: 55

{"method":"Player.GetActivePlayers","params":[],"id":1}
```

A proper jsonrpc 2.0 message MUST include the jsonrpc field (and have a value of `"2.0"`) according to http://www.jsonrpc.org/specification

That is, the proper request should be:

    {"jsonrpc":"2.0","method":"Player.GetActivePlayers","params":[],"id":1}

I haven't figured out enough of rust/cargo to submit a patch yet though.

apoelstra:
Fixed by https://github.com/apoelstra/rust-jsonrpc/issues/3

ecksun:
Alright, I just realized fixing the response. I have been up and down serde trying to figure out why it couldn't parse the response, and its obviously because the Response object doesn't have the jsonrpc field either.

I will hopefully soon submit a patch which addresses that as well.

PR #3, opened 2015-10-25
Add jsonrpc field and always set it to "2.0"

ecksun:
This is one way of solving issue #2 .

While it works, I haven't added any tests, I'm simply not experienced enough with rust to start messing with mocks and such, which I think is required to make a good tests for this.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -85,7 +85,8 @@ impl Client {
         Request {
             method: name,
             params: params,
-            id: JsonValue::U64(*nonce)
+            id: JsonValue::U64(*nonce),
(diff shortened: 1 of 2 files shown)


apoelstra:
Hi @ecksun, welcome to the Rust community! Sorry for the delay in dealing with this.

Right now we don't have a server interface (this is a TODO; if anyone wants to take it they're welcome to), so it's hard to mock things directly. So for now I think the existing serialization round-trip tests are sufficient. Once we have a server we can think about proper test infrastructure.

So I'm going to accept this as-is.

Thanks for your contribution!

PR #4, opened 2015-11-01
Include jsonrpc field in Response

ecksun:
This is required per the jsonrpc v2 specification

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,7 +61,9 @@ pub struct Response {
     /// An error if there is one, or null
     pub error: Option<error::RpcError>,
     /// Identifier from the request
-    pub id: json::Value
+    pub id: json::Value,
+    /// jsonrpc field, MUST be "2.0"
+    pub jsonrpc: json::Value
 }
 
 impl Response {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -61,7 +61,9 @@ pub struct Response {
     /// An error if there is one, or null
     pub error: Option<error::RpcError>,
     /// Identifier from the request
-    pub id: json::Value
+    pub id: json::Value,
+    /// jsonrpc field, MUST be "2.0"
+    pub jsonrpc: json::Value
```
apoelstra: Can you please make this an `Option<json::Value>`? Otherwise there will be parse errors with RPC clients that don't include it.

I know this violates the spec, but right now we're capable of parsing JSONRPC 1.0 and I'd like to keep that.

ecksun: yeah, sure, should I make it optional for the request to then?

I think it might be a better idea to separate the two implementations though, thats at least what I have found to be common when working with jsonrpc implementations.

apoelstra: My feeling is that it's fine if it's mandatory on the request. JSONRPC 1 clients will just ignore the extra field.

ecksun:
Ping!

Does this look alright or do you want further changes?

apoelstra:
Heya, sorry, I will try to get to this by end-of-day.

apoelstra:
This is great, thanks! Sorry for the delay.

PR #8, opened 2018-07-25
Upgrade serde to 1.0.

### LINKED ISSUE #6: Update for serde 0.7

jeandudey:
Fixes #6


dpc:
strason was explicitly about preserving order of keys, allowing duplicates and not parsing numbers. Is this neccessary for jsonrpc?

jeandudey:
@dpc, It isn't strictly necessary for jsonrpc but necessary for bitcoin rpc related crates, serde_json doesn't parse numbers with the `arbitrary_precision` feaure and preserves order with the `preserve_order`, however I don't know if it allows duplicates.

apoelstra:
I'm going to see if I can update `strason` to use serde 1.0.

I upgraded rust-secp256k1 to serde 1.0 yesterday and have some new ideas about how to preserve strason `Json` objects through round trips.

jeandudey:
Is there any particular reason to prefer `strason` over just using `serde_json`?, I have been trying to rewrite `strason` to use `serde` 1.0 and the more I progress the more I'm convinced that it's just a clone of `serde_json` but without the same API, the features `strason` offered are now actually implemented by `serde_json` only missing duplicate fields feature (Is it needed at all for Bitcoin JSON-RPC?).

With the `arbitrary_precision` feature `serde_json` uses a similar approach (but without `unsafe`) that `strason` uses to not lose precision on integers. [1](https://github.com/serde-rs/json/blob/2a62402b184dc2c64d01cca1b3ff72fef3124ae1/src/de.rs#L109)

Using `serde_json` is the better option since it's a lot more tested and reviewed, with also high quality code.

apoelstra:
I've updated strason to serde 1.0; the new version is 0.4.0. It is not quite a drop-in replacement but should be close.

apoelstra:
@jeandudey serde_json does not compile with rust 1.14. See https://github.com/serde-rs/json/pull/332

jeandudey:
@apoelstra I've updated the code to use strason and compiles fines on stable, but on 1.15.0 there are problems with hyper dependencies that don't build.

jeandudey:
Also I have updated `.travis.yml` to build against 1.15.0

apoelstra:
Thanks! Very frustrating about 1.14.0. See https://github.com/servo/unicode-bidi/issues/46 for some history about this.

For now we might have to bump the minimum compiler version for rust-jsonrpc to 1.17. (Which I'm aware brings us down to zero reasons to prefer strason over serde_json -- except that I hope with newer HTTP client dependencies we can get back to 1.14 in the near future.)

apoelstra:
utACK.

Master is not building for me right now because an openssl dep was introduced at some point (wtf, I'm sure I disabled this, but I guess something re-added it in a patch version), and openssl-sys does not even compile anymore. I'll need to fix that in another PR -- will also add travis support and maybe some fuzzers to bring this project more in line with rust-bitcoin norms.

apoelstra:
Never mind - `ssl` is just on by default in this project. I may change the default, especially since our version openssl-sys is broken, but it's not open.

apoelstra:
ACK. The actual minimum compiler version appears to be 1.21 right now, because the rust-url folks have continued increasing their compiler version. But I'll merge this and publish 0.8 so that a latest-serde version is available, and defer fixing `hyper` to a later rev.

alexreg:
Nice work on this. It's great that things have finally moved over to serde, which establishes a consensus. @apoelstra can then take advantage of all the other great features of serde in the future. I'm curious though, is there a reason strason still even exists as a library?

apoelstra:
@alexreg it is the only json parsing library that compiles with Rust 1.14, which is the currently shipping version of rustc in Debian stable on several architectures.

alexreg:
@apoelstra Oh I see, that problem still. That's an issue with Debian of course, but fair enough. So does your library switch to use serde when available, or what?

alexreg:
@apoelstra Hmm, what do HTTP libraries have to do with anything? I'm not sure I understand the first point.

apoelstra:
@alexreg rust-jsonrpc implements a HTTP jsonrpc client.

alexreg:
Sure, but how is that related to what serialisation library is used?

apoelstra:
@alexreg if we can use `serde_json`, we'd rather use `serde_json`. And if we're not supporting 1.14 (and rust-jsonrpc is not, due to lack of HTTP libs), we can use `serde_json`.

alexreg:
Okay, I see. Very good. So the `strason` dependencies for `rust-jsonrpc` and `rust-bitcoin` will both disappear in time?

apoelstra:
Yes.

alexreg:
Great. Sorry for the confusion.

PR #10, opened 2018-07-30
disable ssl by default

apoelstra:


apoelstra:
Running this through a PR so there is a record on github, but this is necessary to publish 0.8 since the build does not succeed on my machine otherwise.

Issue #15, opened 2018-10-29
How to handle `result: null`?


dpc:
For bitcoind, if I call `invalidateblock`, the `result` is `null`:

```
curl --user user  --data-binary '{"jsonrpc": "2.0", "id":"curltest", "method": "invalidateblock", "params": ["0d99e36fe7b3a91b01dc393da7c5e900c17a28390a4ba60820eaa51af60a029e"] }' -H 'content-type: text/plain;' http://127.0.0.1:45929/
Enter host password for user 'user':
{"result":null,"error":null,"id":"curltest"}
```
Make no mistake. The operation was successful.

However it turns into:

```
Response: Err(NoErrorOrResult

Response format is invalid)
```

when calling `into_result`. 

https://www.jsonrpc.org/specification#response_object states:
```
result
    This member is REQUIRED on success.
    This member MUST NOT exist if there was an error invoking the method.
```

I'm not sure if `result: null` is conforming to the standard  or not, but I side with "yes". There is no result to return, so it's `null`.

Does it mean that `into_resul()` is handling this case incorectly, or should I be just not calling `into_result()`? It seems to me that it would be better if it wasn't neccessary to call different methods, depending on the expected result (or lack of it).

dpc:
Ping @jeandudey as bitcoin-rpc library is affected (I'll try to submit a PR), as I'm working on it.

apoelstra:
Thanks Dawid!,

I agree that Core's behaviour here is perfectly reasonable. Unfortunately serde(-json) makes it hard to distinguish nulls from missing fields.

The simplest solution may be to just remove this error path and accept that nonconforming implementations that fail to provide either field will get away with it.



On Sun, Oct 28, 2018 at 11:42:45PM -0700, Dawid Ciężarkiewicz wrote:
> Ping @jeandudey as bitcoin-rpc library is affected (I'll try to submit a PR), as I'm working on it.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/apoelstra/rust-jsonrpc/issues/15#issuecomment-433803518

-- 
Andrew Poelstra
Research Director, Mathematics Department, Blockstream
Email: apoelstra at wpsoftware.net
Web:   https://www.wpsoftware.net/andrew

"Make it stop, my love; we were wrong to try
 Never saw what we could unravel in traveling light
 Nor how the trip debrides like a stack of slides
[...]

PR #16, opened 2018-10-30
Null result

dpc:


dpc:
@stevenroose It's not like this library will generate the second version. It's just that it will accept it as "success" (since it doesn't have `error` field anyway). It the sender that is at fault, and we will just go along and consider it "success", because we can't easily know if it was first or second version at serde level. There's nothing in the spec that says receiver must handle misbehaving sender one way or another, so technically we're good. And without this change some some Bitcoin RPCs are just not working (`invalidateblock`), so we have to do something.

stevenroose:
`invalidateblock` returns no result? *sigh* 

Yeah there is a lot of calls in it's API that should produce empty results.

dpc:
I forgot to link to the issue: https://github.com/apoelstra/rust-jsonrpc/issues/15

apoelstra:
Needs rebase. But concept ACK -- I agree this is kinda inelegant but that's the nature of jsonrpc.

dpc:
Done

dpc:
Hello? :)

PR #18, opened 2018-11-03
Add .rustfmt.toml and .editorconfig

stevenroose:
Makes it a lot easier for contributes to avoid accidentally introducing random changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+# see https://editorconfig.org for more options, and setup instructions for yours editor
+
+[*]
+indent_style = space
+indent_size = 4
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+use_small_heuristics = "Off"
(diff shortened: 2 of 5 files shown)


apoelstra:
We are not using rustfmt before 1.0 and we are *definitely* not overriding the tab settings to disagree with every other rust-bitcoin project and the wider Rust community for 4 years.

stevenroose:
https://github.com/rust-bitcoin/rust-lightning

> Notes on coding style:
> 
>  - Use tabs. If you want to align lines, use spaces. Any desired alignment should display fine at any tab-length display setting.

Whoops. Fair enough. Can change to spaces. It's just really annoying to have to worry about code formatting while programming. It's a delight to be able to ignore format-related typos and have them corrected on `:w` ;)

apoelstra:
Heh, fair enough. In fairness please do not copy Matt's format style to any other projects ;)

apoelstra:
While you're at it can you add a commit that replaces `try!` with `?`? Can rustfmt enforce that?

stevenroose:
I could, but fwiw that block gets removed in #20 and isn't really in the scope of this change. rustfmt doesn't do that. Not sure why. I guess it should only do formatting.

apoelstra:
Ok, sgtm

PR #19, opened 2018-11-04
Some ideas for improvement

dpc:
While `Request` and `Response` might be useful, for a lot of use-cases they are just an implementation detail, that just gets in the way. I was wondering if they could be made non-private altogether, but I guess for some advanced uses, they might be necessary.

So the first patch makes some allocation savings, and second add a convenience function that allows forgetting that `Response` and `Request` exist at all. First patch have a longer commit explaining more about what's going on.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -25,6 +25,7 @@ use std::sync::{Arc, Mutex};
 use hyper;
 use hyper::client::Client as HyperClient;
 use hyper::header::{Authorization, Basic, Headers};
+use serde;
 use serde_json;
 
 use super::{Request, Response};
@@ -61,8 +62,7 @@ impl Client {
         args: &[serde_json::value::Value],
     ) -> Result<T, Error> {
         let request = self.build_request(rpc_name, args);
-        let response = self
-            .send_request(&request)?;
+        let response = self.send_request(&request)?;
 
         Ok(response.clone().into_result()?)
     }
--- a/src/client.rs
+++ b/src/client.rs
@@ -83,14 +83,13 @@ impl Client {
                 if e.kind() == io::ErrorKind::BrokenPipe
                     || e.kind() == io::ErrorKind::ConnectionAborted
                 {
-                    try!(
-                        self.client
-                            .post(&self.url)
-                            .headers(retry_headers)
-                            .body(&request_raw[..])
-                            .send()
-                            .map_err(Error::Hyper)
-                    )
+                    try!(self
+                        .client
+                        .post(&self.url)
+                        .headers(retry_headers)
+                        .body(&request_raw[..])
+                        .send()
+                        .map_err(Error::Hyper))
                 } else {
                     return Err(Error::Hyper(hyper::error::Error::Io(e)));
                 }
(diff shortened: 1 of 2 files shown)


dpc:
This is how the user API changes: https://github.com/dpc/rust-bitcoin-rpc/compare/dpc-dev...dpc:dpc-dev-jsonrpc-reorg

apoelstra:
I like this, though I wonder if it's possible to go farther. At Blockstream we have a similar wrapper to your `do_rpc` function, but with a couple more changes:

1. We have a `do_rpc_hex` function which additionally hex-decodes the data and then passes the result to rust-bitcoin's `deserialize` function. This probably doesn't make sense because it'd put a rust-bitcoin dependency on rust-jsonrpc, but it'd be cool if there was a way to do this. This is useful for RPCs like `getblock` and `getrawtransaction` which support dumping raw hex-encoded data and evading json altogether.
2. We use the RPC `getblockchaininfo` call to determine whether a bitcoind has started up. When it first turns on it returns the error code -9. After that it will return a valid `getblockchaininfo` output. This requires using the Request/Response distinction. Do you think there's a way we can make that more ergonomic?

dpc:
1. Yeah. [This is a common pattern in in rust-bitcoin-rpc too](https://github.com/dpc/rust-bitcoin-rpc/blob/8fcb2e8ed73c539ae9c8e3688efdb6b88ca4720f/src/lib.rs#L276). Because it depends on `rust-bitcoin` I'd just generalize it exactly like you described and put in soon to be a `rust-bitcoincore-rpc`.

2. I was actually doing something lamer. I belive this also belongs to `rust-bitcoincore-rpc`. I would actually make it something higher-level like `fn wait_for_node_to_be_ready(&self) -> Result<()>` so users don't have to know the gnarly details of "error code -28".

dpc:
1. [like this](https://github.com/dpc/rust-bitcoin-rpc/commit/c6a5b2f33f2e9c5b18fe977bbc5889a12ef6cfc8)?

src/client.rs
```diff
@@ -54,6 +55,18 @@ impl Client {
...
+    /// Make a request and deserialize the response
+    pub fn do_rpc<T: for<'a> serde::de::Deserialize<'a>>(
+        &self,
+        rpc_name: &str,
+        args: &[serde_json::value::Value],
+    ) -> Result<T, Error> {
+        let request = self.build_request(rpc_name, args);
+        let response = self.send_request(&request)?;
+
+        Ok(response.clone().into_result()?)
```
stevenroose: Why is this clone required here?

dpc: I think it isn't! :D

dpc: Gone. Thank you!

stevenroose:
ACK

src/client.rs
```diff
@@ -114,14 +113,14 @@ impl Client {
     }
 
     /// Builds a request
-    pub fn build_request(&self, name: String, params: Vec<serde_json::Value>) -> Request {
+    pub fn build_request<'a>(&self, name: &'a str, params: &'a [serde_json::Value]) -> Request<'a> {
```
apoelstra: Can you split this into two lifetimes, one for `name` and one for `params`?

dpc: Done.

apoelstra:
ACK

PR #22, opened 2018-12-19
WIP: Basic no-deps HTTP client for bitcoind RPC

sgeisler:
This PR builds on top of @stevenroose's PR #20. It implements `HttpRoundTripper` for a very basic HTTP client that probably only works with `bitcoind`. I tested it locally with a `bitcoind v0.16.3` using https://github.com/sgeisler/rust-bitcoincore-rpc/commit/2f0d4dbb6a6b627a601cb5cdc41197cc8ebb91b1.

If you think this goes into the right direction I'll add integration tests with some bitcoin core versions.


stevenroose:
Not sure the jsonrpc crate is the right place to add this. We just decided
to take out the HTTP client and make it protocol only.

Could be a separate crate, or function as the default client in
bitcoincore-rpc. Especially if you say it'll only work with bitcoind, I
think it shouldn't be here.

sgeisler:
@stevenroose makes sense, I manly wanted to gather some feedback here, but I'll close this PR and create a new repo once your PR is merged/released.

apoelstra:
Replaced by #34

PR #24, opened 2019-03-22
Support JSON-RPC batching

stevenroose:
One drawback about this implementation is that the responses get lost when a batch error is produced. If retaining those is desired, suggestions are welcome.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -18,6 +18,7 @@
 //! and parsing responses
 //!
 
+use std::collections::HashMap;
 use std::io;
 use std::io::Read;
 use std::sync::{Arc, Mutex};
@@ -29,6 +30,7 @@ use serde;
 use serde_json;
 
 use super::{Request, Response};
+use util::HashableValue;
 use error::Error;
 
 /// A handle to a remote JSONRPC server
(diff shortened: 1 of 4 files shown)


stevenroose:
Made it compile for Rust 1.24.0.

@apoelstra since this doesn't change any existing API, could it be merged faster than the client-agnostic PR? :)

src/client.rs
```diff
@@ -126,6 +136,48 @@ impl Client {
...
+                    matching = Some(i);
+                    break;
+                }
+            }
+
+            match matching {
+                Some(m) => results.push(Some(responses.remove(m))),
+                None => results.push(None),
+            }
+        }
```
apoelstra: Yikes :) this is O(n^2).

Can you instead use an algorithm where you
1. Loop over all responses, putting them in an id→Response `HashMap`,
2. To generate the returned array, loop over all requests, removing objects from the `HashMap` as you go?

stevenroose: Hmm, yeah did this making the assumption that most servers would just order them correct anyway, making this O(n). But I guess we shouldn't make that assumption.

stevenroose: Please review the updated code.

src/client.rs
```diff
@@ -126,6 +137,45 @@ impl Client {
...
+        let responses: Vec<Response> = self.send_raw(&requests)?;
+        if responses.len() > requests.len() {
+            return Err(Error::WrongBatchResponseSize);
+        }
+
+        // First index responses by ID and catch duplicate IDs.
+        let mut resp_by_id = HashMap::new();
+        for resp in responses.into_iter() {
+            // IDs can only be of JSON type String or Number (or Null),
+            // so converting to string should be inexpensive.
```
apoelstra: FYI it's guaranteed to result in an allocation/deallocation pair, which could fragment memory and be very expensive indeed.

src/client.rs
```diff
@@ -126,6 +137,45 @@ impl Client {
...
+            if let Some(dup) = resp_by_id.insert(resp.id.to_string(), resp) {
+                return Err(Error::BatchDuplicateResponseId(dup.id));
+            }
+        }
+        // Match responses to the requests.
+        let results =
+            requests.into_iter().map(|r| resp_by_id.remove(&r.id.to_string())).collect();
```
apoelstra: Oof, and *another* string conversion.

src/client.rs
apoelstra: Comment should be updated now to "cloning should be inexpensive and require no allocations".

PR #31, opened 2019-09-27
client: add socks5 support by passing in HyperClient

jb55:
This will allow us to use a socks5 (or any) connection.

I'm currently looking into adding socks5 proxy support for rust-bitcoincore-rpc. I found the [hyper-socks](https://github.com/sfackler/rust-hyper-socks/blob/master/src/lib.rs#L203) library for creating a socks5 HyperClient. It looks like they create a new HyperClient with a socks5 connector. I just need a way to pass this to the jsonrpc Client.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -43,8 +43,9 @@ pub struct Client {
 }
 
 impl Client {
-    /// Creates a new client
-    pub fn new(url: String, user: Option<String>, pass: Option<String>) -> Client {
+    /// Creates a new client with a specific HyperClient
+    /// Use this to create a socks5 client, etc
+    pub fn with_client(url: String, user: Option<String>, pass: Option<String>, client: HyperClient) -> Client {
         // Check that if we have a password, we have a username; other way around is ok
         debug_assert!(pass.is_none() || user.is_some());
 
@@ -52,9 +53,14 @@ impl Client {
             url: url,
             user: user,
             pass: pass,
-            client: HyperClient::new(),
+            client: client,
             nonce: Arc::new(Mutex::new(0)),
         }
+    }
+
+    /// Creates a new client
+    pub fn new(url: String, user: Option<String>, pass: Option<String>) -> Client {
+        Client::with_client(url, user, pass, HyperClient::new())
     }
 
     /// Make a request and deserialize the response


dpc:
AFAIR, we want to go #20  so anyone will be able to use any client they want.

jb55:
@dpc ah I didn't see that one! yeah that PR is a much nicer solution. Although at first glance it looks like it would make a socks5 implementation more difficult since I only found the hyper-socks one :(

jb55:
oh nm it should still work via https://github.com/stevenroose/rust-jsonrpc-hyper ...

shesek:
Any chance to revisit this? I would very much like to see this implemented, tor support is one of the top requests I'm getting for [bwt](https://github.com/shesek/bwt) (which relies on rust-jsonrpc via rust-bitcoincore-rpc).

@justinmoon even has a working [rust-hyper-socks fork](https://github.com/justinmoon/rust-hyper-socks) that could be used with this.

jb55:
#20 is a larger change, so it might make sense to do this in the meantime? It's a simple way to get a socks5 working.

dpc:
I made a ping in https://github.com/apoelstra/rust-jsonrpc/pull/20 . Please give your gentle reminders if you're blocked on it. :)

jb55:
for reference, here's the code I uses with this branch:

(pseudo-ish code, I copy pasted this and probably needs a few `?`s)

```rust
use hyper::client::Client as HyperClient;
use hyper_socks::Socks5HttpConnector;

let connector = Socks5HttpConnector::new(url)
let socks5_client = connector.map(hyper::Client::with_connector)

let jsonrpc = RpcClient::with_client(
    rpc_url.to_string(),
    Some(rpc_user),
    Some(rpc_pass),
    socks5_client,
);

```

not as general as #20, but a lot less work if you're already using hyper.

apoelstra:
I'm going to close this because we've made a few massive changes in the interim:

* Adding pluggable backends and our own `simple_http` basic-HTTP-client-in-one-file implementation
* Added a `proxy` feature flag for `simple_http` to enable a socks5 proxy.

We could definitely clean that up, and maybe make it generic rather than feature-flagged, but for now I think we've covered the spirit of this PR.

PR #32, opened 2019-11-14
Switch to AtomicU64 counter

hlb8122:
**Justification**

- `AtomicU64` has better performance than `Arc<Mutex<u64>>`.
- The `fetch_add` fits nicely here.
- `last_nonce` doesn't really make sense here, because you get `0` just after initializing the client (before you have used any nonce). `next_nonce` however breaks the API.

**Notes**
- `AtomicUsize` is more portable. https://doc.rust-lang.org/stable/std/sync/atomic/index.html#portability

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -21,7 +21,7 @@
 use std::collections::HashMap;
 use std::io;
 use std::io::Read;
-use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicU64, Ordering};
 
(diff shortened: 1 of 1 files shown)


apoelstra:
Curious about the motivation for closing

hlb8122:
@apoelstra No idea, I forgot I had this PR when I deleted the branch - perhaps that did it. I can restore it if you'd like.

PR #33, opened 2020-08-07
add MSRV details to README

apoelstra:


stevenroose:
I'm getting an error on the last command for v1.22:

```
$ rustup run 1.22.0  cargo update -p 'log:0.4.11' --precise '0.4.8'
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: no matching version `= 0.3.9` found for package `log` (required by `hyper`)
location searched: registry https://github.com/rust-lang/crates.io-index
versions found: 0.4.8
```

While doing that same `cargo update` using stable works. But I assume users with 1.22 can't just use stable to apply their cargo updates..

apoelstra:
That line is frustratingly fragile. Unless you have `0.4.11` installed it just won't work, but I can't figure out how to make it more general.

My guess is that it isn't actually needed for 1.22?

apoelstra:
Let's just bump the MSRV to 1.29 so we can drop this line.

apoelstra:
@stevenroose how about now?

apoelstra:
Needs update in light of #20 (should match the contrib/test.sh commands from #34)

stevenroose (approved):

PR #34, opened 2020-09-26
add basic `HttpRoundTripper` implementation; minor cleanups

apoelstra:
Takes @sgeisler implementation from #22


src/bitcoind_client.rs
```diff
@@ -0,0 +1,188 @@
...
+        SimpleBitcoindClientBuilder::new()
+    }
+}
+
+/// Try to read a line from a buffered reader. If no line can be read till the deadline is reached
+/// return a timeout error.
+fn get_line<R: BufRead>(reader: &mut R, deadline: Instant) -> Result<String, Error> {
+    let mut line = String::new();
+    while deadline > Instant::now() {
+        match reader.read_line(&mut line) {
```
dpc, line 80: This only works if `reader` is non-blocking? And I can't really tell where in `SimpleBitcoindClient::post` `TcpSocket` is set to non-blocking.

apoelstra: Good catch. Yeah, this method looks really weird -- the call to `yield_now` suggests this is supposed to be run in its own thread, but even then, a blocking socket means that the `while` potentially won't ever be reached. There is also a bunch of timeout logic in the code which is used to set `deadline` but which never affects the underlying socket.

Let me see if I can add some unit tests for this.

apoelstra: Lol, upon actually testing this there are a ton of ergonomic issues with the API. I will add some more commits.

apoelstra: So, for what it's worth, the timeout logic seems to work when testing. I'm just not sure how :)

apoelstra:
Added a bunch of cleanup commits. I have a unit test locally which talks to my bitcoind but I'm not sure how best to commit it, since it currently has my local RPC credentials hardcoded and naturally that won't work for anybody else.

src/bitcoind_client.rs
apoelstra: @sgeisler can you offer some insight as to why the timeout code works? I'm really at a loss here, after reading various docs and some of the libstd source I don't see why the first call to `read_line` won't block indefinitely.

dpc: Is some call to https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_nonblocking hiding somewhere?

src/simple_rtt.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+        sock.write_all(b"\r\n")?;
+        sock.write_all(request.body())?;
+
+        // Receive response
+        let mut reader = BufReader::new(sock);
+
+        // Parse first HTTP response header line
+        let http_response = get_line(&mut reader, request_deadline)?;
+        if http_response.len() < 12 || !http_response.starts_with("HTTP/1.0 ") {
```
dpc, line 131: Wouldn't `HTTP/1.1` be a better choice? I would imagine that's the most commonly used version, and some servers might just hard-code 1.1 response, instead of downgrading to 1.0 because they don't expect anyone to call with 1.0. Requesting 1.1 and accepting both 1.1 and 1.0 seems like not a lot of complexity with some additional compatibility.

apoelstra: Sure, added a commit (and checked that the Liquid integration tests pass when using this to do jsonrpc)

darosior (approved):
ACK 451f07fefc437215bebe737ccfed016e324e41ad

Tested downstream. Do you plan to do a new release after merging this ?

src/simple_rtt.rs
```diff
@@ -0,0 +1,201 @@
...
+        // Parse first HTTP response header line
+        let http_response = get_line(&mut reader, request_deadline)?;
+        if http_response.len() < 12 || !http_response.starts_with("HTTP/1.1 ") {
+            return Err(Error::HttpParseError);
+        }
+        match http_response[9..12].parse::<u16>() {
+            Ok(200) => {},
+            Ok(e) => return Err(Error::ErrorCode(e)),
+            Err(_) => return Err(Error::HttpParseError),
+        };
```
darosior, line 138: Unfortunately, bitcoind will (most of the time...) return both an HTTP and a JSONRPC error on a failed call. However, an error can sometimes be expected and no matter if it was it's always useful to get the JSONRPC as well. So, could we instead read the message till the end even if `!= 200` here and hand it to the caller so that it can decide if it should fail, log, or if it was an expected failure ?

apoelstra:
Yes, the plan is to do a release after merge if I can get a signoff from @stevenroose

apoelstra:
@darosior I added a commit which does not error on a bad HTTP status code. Is this what you were thinking?

I could alternately try to parse the response body, and failing that, return the status code if it was non-200. Rather than always ignoring it.

darosior:
> @darosior I added a commit which does not error on a bad HTTP status code. Is this what you were thinking?
> I could alternately try to parse the response body, and failing that, return the status code if it was non-200. Rather than always ignoring it.

Yeah, i was thinking about the latter but will try out anyways, thanks!

apoelstra:
Ok, added a commit which does the latter.

Also removed a panic which had no right to be there

darosior:
Great, thanks for doing this:
```
[2020-11-09][21:53:21][revaultd::bitcoind::actions][ERROR] Error checking bitcoind: Bitcoind error: Making request: RPC error response: RpcError { code: -28, message: "Loading block index...", data: None }
```
Where i previously had something like "500 internal server error" :D

ACK  328a6f3

src/simple_rtt.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+    Err(Error::Timeout)
+}
+
+impl HttpRoundTripper for Tripper {
+    type ResponseBody = Cursor<Vec<u8>>;
+    type Err = Error;
+
+    fn post(&self, request: http::Request<&[u8]>) -> Result<(http::Response<Self::ResponseBody>, u16), Self::Err> {
```
dpc, line 97: https://docs.rs/http/0.2.1/http/status/struct.StatusCode.html instead of `u16`?

src/simple_rtt.rs
```diff
@@ -0,0 +1,198 @@
...
+        Builder::new()
+    }
+}
+
+/// Try to read a line from a buffered reader. If no line can be read till the deadline is reached
+/// return a timeout error.
+fn get_line<R: BufRead>(reader: &mut R, deadline: Instant) -> Result<String, Error> {
+    let mut line = String::new();
+    while deadline > Instant::now() {
+        match reader.read_line(&mut line) {
```
dpc, line 81: Isn't `read_line` blocking? I have a deja vu. :D

apoelstra: lol, that's right, we never did get to the bottom of this

sgeisler: Sorry I totally forgot about that PR. Some months ago I apparently wrote an answer but did it as a review comment accidentally, so GH never published it :smile: 

> I don't think that works at all in the general case :see_no_evil: even with the underlying stream being nonblocking, the buffered reader would still block until a newline is hit. I never really completed that PR, the main use case was to detect core answering correctly but much to slow (idk why tbh), assuming it always sends complete lines at a time. It's stupid. Nowadays I'd probably just use `async` and timeouts. Or if not possible maybe that would require a custom buffered reader based on a nonblocking underlying stream? The `thread::yield` hack is rather horrible though imo.

> EDIT: I remember, the problem was that core was answering to slowly and I had to wait for lines to arrive (was returning `Ok(0)` a lot) so I built this helper fn, but apparently thought it would be clever to have an option to give up after some time. But it's super unreliable as it assumes Core to send full lines quickly and just take time between them.

apoelstra: This is even more confusing :). It **does** work, even with a blocking stream and even with `BufRead` adding blocking on top of that blocking. And I didn't notice any slowness, though to be honest I didn't benchmark either.

apoelstra:
Squashed into #36. Further comments should be filed as issues.

PR #36, opened 2020-12-14
Make HTTP-agnostic and incorporate simple HTTP transport

stevenroose:
- Changes the HttpRoundtripper into Transport which take serializable data and returns a deserializable. This enables usage of JSON-RPC over non-HTTP transports such as unix sockets (like clightning does).
- Takes the base-minimum HTTP client from https://github.com/apoelstra/rust-jsonrpc/pull/34 and ports it to the new trait.
- Adds a `Client::simple_http` constructor that creates a new jsonrpc client using the http transport.

The idea is that other transports can be added with feature flags so that easy constructors can be added for those transports. F.e. an optional dependency on `hyper` and a constructor `Client::hyper`. So that users don't need to worry about transports.


src/client.rs
```diff
@@ -18,119 +18,64 @@
...
-pub trait HttpRoundTripper {
-    /// The type of the http::Response body.
-    type ResponseBody: io::Read;
-    /// The type for errors generated by the roundtripper.
-    type Err: error::Error;
+/// An interface for a transport over which to use the JSONRPC protocol.
+pub trait Transport {
+    /// The Error type for this transport.
+    /// Errors will get converted into Box<std::error::Error> so the
+    /// type here is not use any further.
```
apoelstra, line 37: This is a confusing comment and can probably be dropped.

src/client.rs
```diff
@@ -18,119 +18,64 @@
...
-        if let Some(ref user) = self.user {
-            let mut auth = user.clone();
-            auth.push(':');
-            if let Some(ref pass) = self.pass {
-                auth.push_str(&pass[..]);
-            }
-            let value = format!("Basic {}", &base64::encode(auth.as_bytes()));
-            request_builder.header("Authorization", value);
+        method: &'a str,
+        params: &'a [Box<serde_json::value::RawValue>],
```
apoelstra, line 67: why the box?

apoelstra: ok, i see, after reading the `RawValue` docs

apoelstra (approved):
ack 0496b1b25ca22b8f948a66375ff77693a456ed97

apoelstra (approved):
ack c15d2051e3017d0a26c9da04ffeadf16a00caa27

src/client.rs
dpc: FWIW, I think I understood it on a first read.

src/client.rs
```diff
@@ -139,89 +87,88 @@ impl<Rt: HttpRoundTripper + 'static> Client<Rt> {
...
+            .map_err(|e| Error::Transport(e.into()))?;
         if responses.len() > requests.len() {
             return Err(Error::WrongBatchResponseSize);
         }
 
-        // To prevent having to clone responses, we first copy all the IDs so we can reference
-        // them easily. IDs can only be of JSON type String or Number (or Null), so cloning
-        // should be inexpensive and require no allocations as Numbers are more common.
-        let ids: Vec<serde_json::Value> = responses.iter().map(|r| r.id.clone()).collect();
+        //TODO(stevenroose) check if the server preserved order to avoid doing the mapping
```
darosior, line 102: From [the specs](https://www.jsonrpc.org/specification#batch):
> The Server MAY process a batch rpc call as a set of concurrent tasks, processing them in any order and with any width of parallelism.
The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object.

src/simple_http.rs
```diff
@@ -0,0 +1,304 @@
...
+        // Open connection
+        let request_deadline = Instant::now() + self.timeout;
+        let mut sock = TcpStream::connect_timeout(&self.addr, self.timeout)?;
+
+        // Send HTTP request
+        sock.write_all(format!("POST {} HTTP/1.1\r\n", self.url).as_bytes())?;
+        // Write headers
+        sock.write_all(b"Content-Type: application/json-rpc\r\n")?;
+        if let Some(ref auth) = self.basic_auth {
+            sock.write_all(b"Authentication: ")?;
```
darosior, line 130: I think this should be `Authorization` ?

darosior: The content-length is also missing and required by bitcoind...

src/simple_http.rs
```diff
@@ -0,0 +1,304 @@
...
+                return Err(Error::InvalidUrl(url));
+            }
+
+            self.tp.addr = match format!("{}:{}", hostname, port).to_socket_addrs()?.next() {
+                Some(a) => a,
+                None => return Err(Error::InvalidUrl(url)),
+            };
+        }
+
+        self.tp.url = url;
```
darosior, line 234: The url should not contain the ip:port part.. (`/`, `/wallet/aa`)

PR #38, opened 2020-12-17
A few fixes to the simple_http transport

darosior:
~~There is still the `Content-length` header that we don't set but i don't have a nice fix..~~

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -27,8 +27,8 @@ use serde;
 use serde_json;
 
 use super::{Request, Response};
-use util::HashableValue;
 use error::Error;
+use util::HashableValue;
 
 /// An interface for a transport over which to use the JSONRPC protocol.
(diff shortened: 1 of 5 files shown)


darosior:
I can remove d8add2a5c4678141798edf24ee36bad941953174 if we don't want it, was really just to not pollute other commits with my editor's `rustfmt` hook..

darosior:
Still have a few crash testing this (but at least i can talk to bitcoind!), investigating..

darosior:
Alright, i think it's working as previously now. Going to run on it downstream for testing..

apoelstra (approved):
ack 80223c3734c5457184c5f0c385e7d3753ca9d024

apoelstra:
Was this entirely pulled into #39?

darosior:
I think so, closing

PR #39, opened 2020-12-18
Fixes to the Transport impl

stevenroose:
I had a bunch of changes when I noticed https://github.com/apoelstra/rust-jsonrpc/pull/38, so I cherry-picked @darosior's changes on top here.

This makes the Transport into a Box so that our client doesn't need to be generic. Because traits with generic methods are not allowed to be boxed, the trait methods had to be changed to `send_request` and `send_batch` with concrete types.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -18,6 +18,7 @@
 //! and parsing responses
 //!
 
+use std::fmt;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::atomic;
@@ -30,32 +31,32 @@ use super::{Request, Response};
 use util::HashableValue;
 use error::Error;
 
-/// An interface for a transport over which to use the JSONRPC protocol.
-pub trait Transport {
-    /// The Error type for this transport.
-    /// Errors will get converted into Box<std::error::Error> so the
-    /// type here is not use any further.
-    type Err: ::std::error::Error;
 
-    /// Make an RPC call over the transport.
-    fn call<R>(&self, impl serde::Serialize) -> Result<R, Self::Err>
-        where R: for<'a> serde::de::Deserialize<'a>;
+/// An interface for a transport over which to use the JSONRPC protocol.
+pub trait Transport: Send + Sync + 'static {
+    /// Send an RPC request over the transport.
+    fn send_request(&self, Request) -> Result<Response, Error>;
+    /// Send a batch of RPC requests over the transport.
+    fn send_batch(&self, &[Request]) -> Result<Vec<Response>, Error>;
+    /// Format the target of this transport.
+    /// I.e. the URL/socket/...
+    fn fmt_target(&self, f: &mut fmt::Formatter) -> fmt::Result;
 }
 
 /// A JSON-RPC client.
 ///
 /// Create a new Client using one of the transport-specific constructors:
 /// - [Client::simple_http] for the built-in bare-minimum HTTP transport
-pub struct Client<T: Transport> {
-    transport: T,
+pub struct Client {
+    pub(crate) transport: Box<dyn Transport>,
     nonce: atomic::AtomicUsize,
 }
 
-impl<T: Transport + 'static> Client<T> {
+impl Client {
     /// Creates a new client with the given transport.
-    pub fn with_transport(transport: T) -> Client<T> {
+    pub fn with_transport<T: Transport>(transport: T) -> Client {
         Client {
-            transport: transport,
+            transport: Box::new(transport),
             nonce: atomic::AtomicUsize::new(1),
         }
     }
@@ -77,8 +78,7 @@ impl<T: Transport + 'static> Client<T> {
 
     /// Sends a request to a client
     pub fn send_request(&self, request: Request) -> Result<Response, Error> {
-        let res: Result<Response, _> = self.transport.call(&request);
-        res.map_err(|e| Error::Transport(e.into()))
+        self.transport.send_request(request)
     }
 
     /// Sends a batch of requests to the client.  The return vector holds the response
@@ -93,8 +93,7 @@ impl<T: Transport + 'static> Client<T> {
 
         // If the request body is invalid JSON, the response is a single response object.
         // We ignore this case since we are confident we are producing valid JSON.
-        let responses: Vec<Response> = self.transport.call(requests)
-            .map_err(|e| Error::Transport(e.into()))?;
+        let responses = self.transport.send_batch(requests)?;
         if responses.len() > requests.len() {
             return Err(Error::WrongBatchResponseSize);
         }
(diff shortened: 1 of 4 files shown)


src/simple_http.rs
```diff
@@ -230,8 +230,16 @@ impl Builder {
...
+            if let Some(slash) = after_scheme.find("/") {
+                (
+                    &after_scheme[0..slash],
+                    // Replace double slashes that might occur;
+                    // this is pretty common when adding paths together.
+                    after_scheme[slash..].replace("//", "/")
+                )
+            } else {
+                (after_scheme, "/".to_owned())
+            }
```
darosior: the url is already checked for doubled-`/` just after the host part. However i'm not sure it's a good idea to sanitize the caller's input: it's reasonable to expect they are giving a decent path, and we might not do the right thing. For example on Windows, which does some nasty stuff with paths.

src/client.rs
```diff
@@ -30,32 +31,32 @@ use super::{Request, Response};
...
-    fn call<R>(&self, impl serde::Serialize) -> Result<R, Self::Err>
-        where R: for<'a> serde::de::Deserialize<'a>;
+pub trait Transport: fmt::Debug + 'static {
+    /// Send an RPC request over the transport.
+    fn send_request(&self, Request) -> Result<Response, Error>;
+    /// Send a batch of RPC requests over the transport.
+    fn send_batch(&self, &[Request]) -> Result<Vec<Response>, Error>;
+	/// Format the target of this transport.
+	/// I.e. the URL/socket/...
+	fn fmt_target(&self, f: &mut fmt::Formatter) -> fmt::Result;
```
darosior, line 43: nit: indent

src/client.rs
```diff
@@ -144,6 +143,14 @@ impl<T: Transport + 'static> Client<T> {
     }
 }
 
+impl fmt::Debug for ::Client {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "jsonrpc::Client(")?;
+		self.transport.fmt_target(f)?;
+		write!(f, ")")
```
darosior, line 150: nit: indent

src/simple_http.rs
```diff
@@ -184,54 +204,49 @@ impl Builder {
...
+            return Err(Error::InvalidUrl(url.to_owned()));
         }
 
-        self.tp.url = url;
+        self.tp.addr = match format!("{}:{}", hostname, port).to_socket_addrs()?.next() {
+            Some(a) => a,
+            None => return Err(Error::InvalidUrl(url.to_owned())),
+        };
+        self.tp.path = path;
+        println!("creating jsonrpc Client with addr {:?} and path {}", self.tp.addr, self.tp.path);
```
darosior, line 270: nit: debug line?

src/client.rs
```diff
@@ -30,32 +31,32 @@ use super::{Request, Response};
...
 }
 
 /// A JSON-RPC client.
 ///
 /// Create a new Client using one of the transport-specific constructors:
 /// - [Client::simple_http] for the built-in bare-minimum HTTP transport
-pub struct Client<T: Transport> {
-    transport: T,
+pub struct Client {
+    pub(crate) transport: Box<dyn Transport>,
```
darosior, line 51: This made it no longer `Send` :(

src/simple_http.rs
stevenroose: Hmm, I agree messing with user input is tricky. Though I think double slashes are valid in paths no? In any way, if you give url `http://localhost:1234//wallet/name`, bitcoind errors. But yeah that might be just something we shouldn't support.

src/client.rs
stevenroose: oh I'll probably have to `:retab` because of borking vim.

src/client.rs
stevenroose: Oh that's because we need to make `Transport: Send` I think. Box should be Send, I discussed that with Andrew.

src/simple_http.rs
darosior: > In any way, if you give url http://localhost:1234//wallet/name, bitcoind errors

I thought
```rust
            let mut split = after_scheme.splitn(2, "/");
            (split.next().unwrap(), format!("/{}", split.next().unwrap_or("")))
```
prevented that but it doesn't if the user provides a `/path/to/wallet` so i now suggest to either silently correct the first character (drop it if it's a `/` already), or document in `path` that we want an "absolute" path or not. The former seems preferable

stevenroose:
Sorry for the bad commit messages :/

darosior (approved):
ACK 9dc0c48

lightly-tested downstream against bitcoind v0.21

src/simple_http.rs
```diff
@@ -49,6 +50,52 @@ impl SimpleHttpTransport {
...
+
+    fn request<R>(&self, req: impl serde::Serialize) -> Result<R, Error>
+        where R: for<'a> serde::de::Deserialize<'a>
+    {
+        // Open connection
+        let request_deadline = Instant::now() + self.timeout;
+        let mut sock = TcpStream::connect_timeout(&self.addr, self.timeout)?;
+
+        // Send HTTP request
+        sock.write_all(format!("POST {} HTTP/1.1\r\n", self.path).as_bytes())?;
```
apoelstra, line 65: Could you break this apart so that it doesn't use `format!`?

src/simple_http.rs
```diff
@@ -184,54 +204,51 @@ impl Builder {
...
+            },
+            None => DEFAULT_PORT,
+        };
+        // make sure we don't have a second colon in this part
+        if split.next().is_some() {
+            return Err(Error::InvalidUrl(url.to_owned()));
         }
 
-        self.tp.url = url;
+        self.tp.addr = match format!("{}:{}", hostname, port).to_socket_addrs()?.next() {
```
apoelstra, line 260: There should be no need for `format!` here, `ToSocketAddrs` is implemented for `(&str, u16)`

src/simple_http.rs
```diff
@@ -184,54 +204,51 @@ impl Builder {
...
-            self.tp.addr = match format!("{}:{}", hostname, port).to_socket_addrs()?.next() {
-                Some(a) => a,
-                None => return Err(Error::InvalidUrl(url)),
-            };
+        };
+        // (3) split off auth part
+        let after_auth = {
+            let mut split = before_path.splitn(2, "@");
+            let s = split.next().unwrap();
+            split.next().unwrap_or(s)
```
apoelstra, line 243: We should probably check that the protocol is "http://" rather than just throwing it away. But this is fine for now.

src/simple_http.rs
```diff
@@ -184,54 +204,51 @@ impl Builder {
     }
 
     /// Set the URL of the server to the transport.
-    pub fn url<U: Into<String>>(mut self, url: U) -> Result<Self, Error> {
-        let url = url.into();
-
+    pub fn url(mut self, url: &str) -> Result<Self, Error> {
```
dpc, line 220: Extract to a separate pure function, returning or values in a struct, and add at least some basic sanity tests.

src/simple_http.rs
stevenroose: ah yeah that makes sense

src/simple_http.rs
stevenroose: it's quite easy to do, I can add that.

src/simple_http.rs
stevenroose: fixing

src/simple_http.rs
stevenroose: There are some basic sanity tests for this using the Builder. But yeah perhaps those can be more complete.

src/lib.rs
```diff
@@ -51,6 +51,24 @@ pub use client::{Client, Transport};
 
 use serde_json::value::RawValue;
 
+/// Shorthand method to convert an argument into a [Box<serde_json::value::RawValue>].
+/// Since serializers rarely fail, it's probably easier to use [arg] instead.
+fn safe_arg<T: serde::Serialize>(arg: T) -> Result<Box<RawValue>, serde_json::Error> {
```
apoelstra, line 56: Missing `pub` and should be called `try_arg` rather than `safe_arg`

darosior (approved):
re-ACK 6c2fc02

apoelstra:
@stevenroose can you remove the `localhast` and google unit tests? Or at least the google one ... it will break offline bisection.

stevenroose:
Pushed a commit removing those tests.

apoelstra:
Can you remove them in the original commit? It will break offline bisection otherwise.

apoelstra:
ack d65fe1c9c5abc46bddaefbd4c6e39518d79673e6

Can you add another commit which does
```diff
diff --git a/src/error.rs b/src/error.rs
index bd390c4..6d2d1ff 100644
--- a/src/error.rs
+++ b/src/error.rs
@@ -27,7 +27,8 @@ use Response;
 #[derive(Debug)]
 pub enum Error {
     /// A transport error
-    Transport(Box<error::Error>),
+    // extra bounds needed for anyhow
+    Transport(Box<dyn error::Error + Send + Sync>),
     /// Json error
     Json(serde_json::Error),
     /// Error response
```

apoelstra (approved):
ack dec87aabd39d5ed6d8ed03439d24e6e27f1ebb19

PR #42, opened 2021-04-24
More transport implementations

darosior:
This implements:
- A synchronous transport over `TcpListener` (i don't know of any application that use this but it was almost free to have)
- A synchronous Unix Domain Socket transport. This is useful for communicating with Bitcoin-related softwares other than `bitcoind`. For example, the [C-lightning daemon](https://github.com/ElementsProject/lightning/) and [Revault wallet daemon](https://github.com/revault/revaultd/) both expose an RPC interface over an UDS. This could allow applications such as https://github.com/laanwj/rust-clightning-rpc or https://github.com/revault/revault-gui to directly use this crate instead of relying on a modified version of it.

Unfortunately, there is a quirk. The "ip" listener from the stdlib works well on both Unix and Windows but the stdlib has no support for UDS on Windows (even though it has supported them for almost 3 years now). Therefore i had to introduce a third party crate which hacks around an `UnixStream` and `UnixListener` for Windows behind a feature flag. It also probably break the MSRV though i'm not sure as i don't have a Windows machine at hand to test that.

### DIFF
--- a/src/simple_tcp.rs
+++ b/src/simple_tcp.rs
@@ -1,0 +1,162 @@
+//! This module implements a synchronous transport over a raw TcpListener.
+
+use std::{fmt, io, net, time};
+
+use serde;
+use serde_json;
+
+use client::Transport;
+use {Request, Response};
+
+/// Error that can occur while using the TCP transport.
+#[derive(Debug)]
+pub enum Error {
+    /// An error occurred on the socket layer
+    SocketError(io::Error),
+    /// We didn't receive a complete response till the deadline ran out
+    Timeout,
+    /// JSON parsing error.
+    Json(serde_json::Error),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        match *self {
+            Error::SocketError(ref e) => write!(f, "Couldn't connect to host: {}", e),
+            Error::Timeout => f.write_str("Didn't receive response data in time, timed out."),
+            Error::Json(ref e) => write!(f, "JSON error: {}", e),
+        }
+    }
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::SocketError(e)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(e: serde_json::Error) -> Self {
+        Error::Json(e)
+    }
+}
+
+impl From<Error> for ::Error {
+    fn from(e: Error) -> ::Error {
+        match e {
+            Error::Json(e) => ::Error::Json(e),
+            e => ::Error::Transport(Box::new(e)),
+        }
+    }
+}
+
+impl ::std::error::Error for Error {}
+
+/// Simple synchronous TCP transport.
+#[derive(Debug, Clone)]
+pub struct TcpTransport {
+    /// The internet socket address to connect to
+    pub addr: net::SocketAddr,
+    /// The read and write timeout to use for this connection
+    pub timeout: Option<time::Duration>,
+}
+
+impl TcpTransport {
+    /// Create a new TcpTransport without timeouts
+    pub fn new(addr: net::SocketAddr) -> TcpTransport {
+        TcpTransport {
+            addr,
+            timeout: None,
+        }
+    }
+
+    fn request<R>(&self, req: impl serde::Serialize) -> Result<R, Error>
+    where
+        R: for<'a> serde::de::Deserialize<'a>,
+    {
+        let mut sock = net::TcpStream::connect(&self.addr)?;
+        sock.set_read_timeout(self.timeout)?;
+        sock.set_write_timeout(self.timeout)?;
+
+        serde_json::to_writer(&mut sock, &req)?;
+
+        // NOTE: we don't check the id there, so it *must* be synchronous
+        let resp: R = serde_json::Deserializer::from_reader(&mut sock)
+            .into_iter()
+            .next()
+            .ok_or(Error::Timeout)??;
+        Ok(resp)
+    }
+}
+
+impl Transport for TcpTransport {
+    fn send_request(&self, req: Request) -> Result<Response, ::Error> {
+        Ok(self.request(req)?)
+    }
+
+    fn send_batch(&self, reqs: &[Request]) -> Result<Vec<Response>, ::Error> {
+        Ok(self.request(reqs)?)
+    }
+
+    fn fmt_target(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.addr)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        io::{Read, Write},
+        thread,
+    };
+
+    use super::*;
+    use Client;
+
+    // Test a dummy request / response over a raw TCP transport
+    #[test]
+    fn sanity_check_tcp_transport() {
+        let addr: net::SocketAddr =
+            net::SocketAddrV4::new(net::Ipv4Addr::new(127, 0, 0, 1), 0).into();
+        let server = net::TcpListener::bind(&addr).unwrap();
+        let addr = server.local_addr().unwrap();
+        let dummy_req = Request {
+            method: "arandommethod",
+            params: &[],
+            id: serde_json::Value::Number(4242242.into()),
+            jsonrpc: Some("2.0".into()),
+        };
+        let dummy_req_ser = serde_json::to_vec(&dummy_req).unwrap();
+        let dummy_resp = Response {
+            result: None,
+            error: None,
+            id: serde_json::Value::Number(4242242.into()),
+            jsonrpc: Some("2.0".into()),
+        };
+        let dummy_resp_ser = serde_json::to_vec(&dummy_resp).unwrap();
+
+        let client_thread = thread::spawn(move || {
+            let transport = TcpTransport {
+                addr,
+                timeout: Some(time::Duration::from_secs(5)),
+            };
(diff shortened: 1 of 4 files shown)


darosior:
Based on #43

darosior:
Rebased to try to kick CI now that #43 is merged

darosior:
Rebased on #44 for the CI fixes

darosior:
Ok, got rid of the `simple_uds_windows` feature to rather make the usage of the dependency on `simple_uds is active + target is windows`. Also this dep breaks MSRV `1.29` but not so badly as it only jumps to `1.30` in this specific combination.

apoelstra:
I think basically all of the whitespace changes that rustfmt has introduced are bad.

dpc:
> I think basically all of the whitespace changes that rustfmt has introduced are bad.

I like all except one. Might be Stockholm syndrome though.

darosior:
The formatting was based on [the `rustfmt` config file you merged](https://github.com/apoelstra/rust-jsonrpc/pull/18) that the CI just started enforcing :)

Anyways, this is not part of this PR and the `rustfmt` check in CI and `rustfmt` commit were drop from in #44

src/simple_tcp.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::{fmt, io, net, time};
+
+use serde;
+use serde_json;
+
+use client::Transport;
+use {Request, Response};
+
+/// Error that can occur while using the TCP transport.
+#[derive(Debug)]
```
dpc, line 13: Maybe `Clone`? Will it work?

src/simple_uds.rs
```diff
@@ -0,0 +1,185 @@
...
+        thread,
+    };
+
+    use super::*;
+    use Client;
+
+    // Test a dummy request / response over an UDS
+    #[test]
+    fn sanity_check_uds_transport() {
+        let socket_path: path::PathBuf = "uds_scratch.socket".into();
```
dpc, line 130: Multiple tests running at the same time can make each other fail here. Not sure if worth fixing, but some randomization (`std::process::id`) would be better.

src/simple_uds.rs
```diff
@@ -0,0 +1,185 @@
...
+        let dummy_resp = Response {
+            result: None,
+            error: None,
+            id: serde_json::Value::Number(111.into()),
+            jsonrpc: Some("2.0".into()),
+        };
+        let dummy_resp_ser = serde_json::to_vec(&dummy_resp).unwrap();
+
+        let cli_socket_path = socket_path.clone();
+        let client_thread = thread::spawn(move || {
```
dpc, line 151: This code is correct (it seems) when the test works OK, but it can act meh in case it actually find some non-trivial bug somewhere. In the best case - it will just leave a detached thread running, which is OKis, I guess, but in the worse, it might hang on the `.join` later, making the user wonder what have happened, and why is not finishing.

The simplest solution is to just bound the loop to some reasonable value. Since the retry delay is fixed, assuming that this thread shouldn't ever need more than 30s will at least guarantee that the test terminates.

The best structure is using `Arc<AtomicBool>` or something like that to signal to the thread loop termination request, and wrapping the join handle into some newtype, that on drop: sets the termination flag and `join` s on the handle. This way the no bug can hang the thread, and no bug can miss joining on it.

darosior: Ah i was going to comment "hehe, it's not" but i actually forgot to set the timeout in the constructor of the transport. With it you wouldn't block for more than *timeout* seconds on `join`.

src/simple_tcp.rs
darosior: Nope :/

src/simple_uds.rs
darosior: Applied, thanks:
```diff
diff --git a/src/simple_uds.rs b/src/simple_uds.rs
index fe0dfb8..9e506f1 100644
--- a/src/simple_uds.rs
+++ b/src/simple_uds.rs
@@ -118,7 +118,7 @@ mod tests {
     use std::{
         fs,
         io::{Read, Write},
-        thread,
+        process, thread,
     };
 
     use super::*;
@@ -127,7 +127,7 @@ mod tests {
     // Test a dummy request / response over an UDS
     #[test]
     fn sanity_check_uds_transport() {
-        let socket_path: path::PathBuf = "uds_scratch.socket".into();
+        let socket_path: path::PathBuf = format!("uds_scratch_{}.socket", process::id()).into();
         // Any leftover?
         fs::remove_file(&socket_path).unwrap_or_else(|_| ());
 ```

File: src/simple_uds.rs (hunk 3 above)
Comment (darosior) [reply src/simple_uds.rs:151]: Added a timeout in the Transport as mentioned above, and removed the loop. Tested by commenting out the response write from the main thread:
```diff
diff --git a/src/simple_tcp.rs b/src/simple_tcp.rs
index 7a6dfdb..1018f14 100644
--- a/src/simple_tcp.rs
+++ b/src/simple_tcp.rs
@@ -137,19 +137,13 @@ mod tests {
         let dummy_resp_ser = serde_json::to_vec(&dummy_resp).unwrap();
 
         let client_thread = thread::spawn(move || {
-            let transport = TcpTransport::new(addr);
+            let transport = TcpTransport {
+                addr,
+                timeout: Some(time::Duration::from_secs(5)),
+            };
             let client = Client::with_transport(transport);
 
-            loop {
-                match client.send_request(dummy_req.clone()) {
-                    Ok(resp) => return resp,
-                    Err(e) => {
-                        // Print so should it fail it can be debugged with --nocapture
-                        eprintln!("Error while sending request: '{:?}'", e);
-                        thread::sleep(time::Duration::from_millis(100));
-                    }
-                }
-            }
+            client.send_request(dummy_req.clone()).unwrap()
         });
 
         let (mut stream, _) = server.accept().unwrap();
@@ -161,8 +155,8 @@ mod tests {
         }
         assert_eq!(recv_req, dummy_req_ser);
 
-        stream.write(&dummy_resp_ser).unwrap();
-        stream.flush().unwrap();
+        //stream.write(&dummy_resp_ser).unwrap();
+        //stream.flush().unwrap();
         let recv_resp = client_thread.join().unwrap();
         assert_eq!(serde_json::to_vec(&recv_resp).unwrap(), dummy_resp_ser);
     }
diff --git a/src/simple_uds.rs b/src/simple_uds.rs
index 9e506f1..69881bd 100644
--- a/src/simple_uds.rs
+++ b/src/simple_uds.rs
@@ -149,19 +149,13 @@ mod tests {
 
         let cli_socket_path = socket_path.clone();
         let client_thread = thread::spawn(move || {
-            let transport = UdsTransport::new(cli_socket_path);
+            let transport = UdsTransport {
+                sockpath: cli_socket_path,
+                timeout: Some(time::Duration::from_secs(5)),
+            };
             let client = Client::with_transport(transport);
 
-            loop {
-                match client.send_request(dummy_req.clone()) {
-                    Ok(resp) => return resp,
-                    Err(e) => {
-                        // Print so should it fail it can be debugged with --nocapture
-                        eprintln!("Error while sending request: '{:?}'", e);
-                        thread::sleep(time::Duration::from_millis(100));
-                    }
-                }
-            }
+            client.send_request(dummy_req.clone()).unwrap()
         });
 
         let (mut stream, _) = server.accept().unwrap();
```

src/simple_uds.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    use super::*;
+    use Client;
+
+    // Test a dummy request / response over an UDS
+    #[test]
+    fn sanity_check_uds_transport() {
+        let socket_path: path::PathBuf = format!("uds_scratch_{}.socket", process::id()).into();
+        // Any leftover?
+        fs::remove_file(&socket_path).unwrap_or_else(|_| ());
```
dpc, line 132: Not important, but I usually just `let _ = failable_operation();` when I want to ignore errors.

dpc (approved):
LGTM, but I'm only a casual lurker.

apoelstra (approved):
utACK. I didn't actually test these but I trust you have an application where you'll notice errors. I did read the code and confirmed that it seems to make sense.

The Windows stuff is a bit unfortunate but I think the solution is reasonably well-encapsulated.

darosior:
So testing this downstream on Windows through Github Actions uncovered that `[target.'cfg(windows)'.features]` wasn't valid, changed it to:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 1acbe89..74e53b1 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,14 +20,10 @@ default = [ "simple_http", "simple_tcp" ]
 simple_http = [ "base64-compat" ]
 # Basic transport over a raw TcpListener
 simple_tcp = []
-# Basic transport over a raw UnixStream
-simple_uds = []
-
-# Activate a third party crate to use UDSs on Windows
-[target.'cfg(windows)'.features]
+# Basic transport over a raw UnixStream. This activates a third party
+# crate to use UDSs on Windows, but will only compile it on Windows.
 simple_uds = ["uds_windows"]
 
-
 [dependencies]
 serde = "1"
 serde_derive = "1"
```

apoelstra (approved):
ACK ddf40d0291602dd07f04199c573df03a7b57b525

Again, with the caveat that I'm trusting that the Windows stuff works.

darosior:
I've dropped the Windows commit. It's ugly, i've no use for it anymore and i couldn't test it properly (only through CI in a convoluted manner).

apoelstra (approved):
ACK e4d2510d52d27ea87a8321707424b3eb561602af

shesek:
> i don't know of any application that use this but it was almost free to have

The Electrum server RPC protocol uses plain JSONRPC over TCP/TLS

PR #54, opened 2022-02-14
Don't attempt to read a body for auth error (401)

devrandom:
Prevents a long timeout followed by an uninformative error message.

See also https://github.com/JeremyRubin/rust-jsonrpc-async/pull/2

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -100,6 +100,11 @@ impl SimpleHttpTransport {
         // Skip response header fields
         while get_line(&mut reader, request_deadline)? != "\r\n" {}
 
+        if response_code == 401 {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK c0878de077e7b33eb5e172de1193caaf8c807f7e

I didn't check the HTTP spec to see if it matches your claim, but if it's causing problems let's fix it.

devrandom:
Sorry, I wasn't clear.  The bodies *are* permitted in the spec, but Bitcoin Core doesn't send a body, which breaks the previous logic.  If the aim is to be spec compliant, a bigger change would be required.

apoelstra:
Great, sounds good!

I don't really care about the spec :) but it's good to have a record here of what our thinking is.

Issue #59, opened 2022-06-09
Competing json rpc implementations


tcharding:
There is another crate that also implements json rpc: https://github.com/thomaseizinger/rust-jsonrpc-client

I am no where near an expert on this sort of thing but I spent an hour or so implementing some basic code to hit bitcoind by way of this crate and also @thomaseizinger's crate. As far as I can tell the biggest differences outside of cosmetics are:

- Thomas' crate supports async
- Thomas's crate allows transport backends to be plugged in where as this one implements its own simple transports 
- This crate has minimal dependencies (only serde stuff) where as the dependency graph of Thomas' is bigger, albeit still lean until one starts adding transport backends

I'm just flagging this since we are considering bringing this crate under the `rust-bitcoin` ecosystem.

apoelstra:
I think, if @thomaseizinger's crate is (a) maintained, (b) willing to define a fixed MSRV (ideally 1.41, at least when no backends are available), and (c) willing to port our "simple HTTP" backend so that it's usable without reqwest or other stuff, then we could deprecate this crate in favor of that one.

thomaseizinger:
> I think, if @thomaseizinger's crate is (a) maintained, (b) willing to define a fixed MSRV (ideally 1.41, at least when no backends are available), and (c) willing to port our "simple HTTP" backend so that it's usable without reqwest or other stuff, then we could deprecate this crate in favor of that one.

a) It is definitely maintained but help is always appreciated.
b) I'll have to check what the current MSRV is but the goal is to keep it as minimal as possible given the dependencies. `syn` and `serde` are kind of essential, we might be able to drop `url` if it causes problems. I am open to make 1.41 work.
c) If we'd want to use a "simple-http" backend, it would be great to release it as another crate and provide support for it similar to all other clients.

apoelstra:
Regarding (c), our "simple http" backend is probably not general enough to be in a separate crate with a useful API. It basically just echoes hardcoded headers into a socket.

thomaseizinger:
> Regarding (c), our "simple http" backend is probably not general enough to be in a separate crate with a useful API. It basically just echoes hardcoded headers into a socket.

Hmm, I see the issue.

For the client to be async, one needs to choose an async TcpSocket: either `tokio`, `async-std` or something else. `async-io` did some innovation here as well I think.

[...]

PR #60, opened 2022-06-14
Add link to async JSONRPC crate

tcharding:
Only the last patch, draft till the other (clippy) ones merge.

After discussion [here](https://github.com/apoelstra/rust-jsonrpc/issues/59) it seems this crate and [this one](https://github.com/thomaseizinger/rust-jsonrpc-client) are not competing but rather serve different purposes. We are all friends here so lets help our users out and point them to @thomaseizinger crate if they are writing async applications.


README.md
```diff
@@ -45,3 +45,10 @@ git config --local core.hooksPath githooks/
 ```
 
 Alternatively add symlinks in your `.git/hooks` directory to any of the githooks we provide.
+
+## Design goals
+
+This library was built with the primary purpose of talking to Bitcoin Core without the need for
+additional dependencies. This means we do not, and likely never will, support async. If you are
+writing an async application you might be interested in this [alternative JSONRPC
+client](https://github.com/thomaseizinger/rust-jsonrpc-client).
```
thomaseizinger, line 54: We should definitely be linking to https://github.com/jeremyrubin/rust-bitcoincore-rpc-async/ as well.

apoelstra:
No opinion from me; this looks fine.

README.md
tcharding: Interesting, the lib you link depends on https://github.com/jeremyrubin/rust-jsonrpc-async/ which is, as far as I can tell, a fork of this repo _but_ it has no attribution and was not forked using github (i.e., the github rebo does not indicate it is a fork). It also appears that the same was done for the [json](https://github.com/JeremyRubin/rust-bitcoincore-rpc-async/tree/main/json) crate, it seems to be forked from `rust-bitocincore-rpc/json`, again with no attribution and no indication on github that its a fork.

Both this crate and `rust-bitcoincore-rpc` are licensed under creative commons so its strictly correct to do all this. It  smells bad though.

With this in mind, I'm not going to explicitly link to that crate.

thomaseizinger: The commit history is intact though and @jeremyrubin is a known member of the community. I doubt that there were any bad intentions in regards to missing attributions but probably just an oversight.

In any case, I don't think it is fair to just link to my crate and not the others when they are perfectly fine alternatives. So if you choose to not link to them, please also don't link to mine :)

tcharding: Fair points, thanks.

JeremyRubin: Not accurate to say there is no attribution:

1) License is intact and unmodified, along with all license headers
2) Cargo.toml lists @apoelstra as a co-author

The reason for not being a github fork is because github forks are second class citizens, and there is no intention of upstreaming the patched version, since async is a stated non-goal of the original crate.

JeremyRubin: note also that some of these libraries also depend on the sapio-bitcoin libraries, not the rust-bitcoin ones.

The sapio-bitcoin library is intended to eventually be upstreamed though (it's only real changes are, IIRC, adding schemars support to types, but that proved a bit controversial to de-fork).

tcharding: Excuse the misunderstanding, in the Australian vernacular 'no worries' means 'you are alright, I rescind my concerns, we are all good now'.

Not that anyone else should act like me but I tend to go overboard on the attribution, I would have expected to see in the readme something like:

"I hacked this together with code for XYZ basically just to add 'feature', I don't plan on pushing the changes back to them since 'feature' is not wanted"

Issue #40, opened 2021-01-14
Cannot generate documentation


vadixidav:
I have been running into issues with the latest `jsonrpc v0.12.0`. Notably, I cannot generate the documentation due to these errors:

```rust
error[E0433]: failed to resolve: could not find `__private` in `_serde`
  --> /home/worleyg/.cargo/registry/src/github.com-1ecc6299db9ec823/jsonrpc-0.12.0/src/lib.rs:79:24
   |
79 | #[derive(Debug, Clone, Deserialize, Serialize)]
   |                        ^^^^^^^^^^^ could not find `__private` in `_serde`
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `__private` in `_serde`
  --> /home/worleyg/.cargo/registry/src/github.com-1ecc6299db9ec823/jsonrpc-0.12.0/src/lib.rs:79:37
   |
79 | #[derive(Debug, Clone, Deserialize, Serialize)]
   |                                     ^^^^^^^^^ could not find `__private` in `_serde`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

However, I can fix this by adding `serde = { version = "1.0.119", features = ["derive"], default-features = false }` to my `Cargo.toml`:

```toml
jsonrpc = { version = "0.12.0", default-features = false }
serde = { version = "1.0.119", features = ["derive"], default-features = false }
```

It seems like `jsonrpc` doesn't enable the `derive` feature of serde, but for some reason documentation is failing to generate unless it is present. I do not know exactly why that is the case, as the crate builds fine, it is just documentation generation that fails.

apoelstra:
I'm kinda surprised that the crate builds without the derive feature. We should just turn it on.

vadixidav:
> I'm kinda surprised that the crate builds without the derive feature. We should just turn it on.

I think the reason it builds is it uses serde_derive (the crate) directly.

apoelstra:
Ah, it may be that nobody does this anymore and it causes weird breakage. Let me PR to change to use the feature.

PR #41, opened 2021-01-15
use serde `derive` feature rather than `serde_derive`

apoelstra:
Fixes #40

PR #61, opened 2022-06-16
Use serde derive feature instead of serde_derive crate

tcharding:
We do not need to use the `serde_derive` crate, there is a serde feature for this.

This is a replacement for #41 to save you doing it apoelstra.

Fix: #40


apoelstra (approved):
ACK fa9baae299a733e7a530a83c38b2b4b6331e6697

Issue #62, opened 2022-06-30
SimpleHttp IPv6 support


tthebst:
https://github.com/apoelstra/rust-jsonrpc/blob/master/src/simple_http.rs#L302

Url parsing currently fails for IPv6 addresses.

My suggestion would be to sth simpler like this. Instead of trying to split the string.

```
// Needs to be of format <host_name>:<port>
let mut addr = match after_auth.to_socket_addrs() {
    Ok(addr) => addr,
    Err(_) => {
        // Try to add fallback port
        format!("{}:{}", after_auth, fallback_port).to_socket_addrs()?
    }
};
```

PR #63, opened 2022-07-02
Add IPv6 support

tthebst:
Resolves: #62

Idea: 
After removing the auth part of we are left with either `<host_name>` or `<host_name_>:<port>` and `ToSocketAddr` [requires]( https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) `&str` to be of form `<host_name_>:<port>`. First `to_socket_addrs()` call only succeeds if `after_auth` is well formed (`<host_name_>:<port>` ). If this fails we try to add the fallback port and call `to_socket_addrs()` again.

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -298,22 +298,17 @@ impl Builder {
             let s = split.next().unwrap();
             split.next().unwrap_or(s)
         };
-        // so now we should have <hostname>:<port> or just <hostname>
-        let mut split = after_auth.split(':');
-        let hostname = split.next().unwrap();
-        let port: u16 = match split.next() {
-            Some(port_str) => match port_str.parse() {
-                Ok(port) => port,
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 21e2f2cd3fc017795de1506475a1bec73973e951

apoelstra:
Thank you! I'll merge once CI passes.

schildbach:
Is this PR already released, also in the form of a Cargo crate?
I'm investigating into https://github.com/romanz/electrs/issues/792 which fails to parse an afact valid IPv6 URL.

tthebst:
@schildbach Yes, it was released afaik with 0.13.0. But the latest 
`rust-bitcoincore-rpc` (0.16.0) still uses 0.12.0. So you need patch the `json-rpc` dependency until a newer `bitcoincore-rpc` version is released.

PR #64, opened 2022-07-21
Improve CI

tcharding:
Improve CI by doing

- Explicitly return 0 from script
- Remove `TOOLCHAIN` env var
- Add `DO_FEATURE_MATRIX` 
- Add `DO_DOCS` and build docs in ci


apoelstra (approved):
ACK b067391cc6eaedd0884e54c6a5184c9009ab8b2a

PR #69, opened 2022-08-17
Support mult-line response in  buffer read.

flomang:
I found this jsonrpc client to be the easiest to understand from a foundational perspective. May we have multi-line response support though? I tried using the original against a grin node and they pretty print their responses thus resulting in a parse exception.


apoelstra (approved):
ACK 6e6355f031fb253106c4d9813054d4652bae1d4f

apoelstra:
This seems like a reasonable enough feature. Even if we only focus on Bitcoin Core it's kinda fragile for us to assume there will never be newlines.

flomang:
Thank you!

Issue #57, opened 2022-06-07
[Feature Request] Enable Socks5 proxy support at Transport level


rajarshimaitra:
We had an usecase in BDK to have core-rpc client talk through a Tor proxy, like the electrum client, via a proxy configuration.

Potential use case: Mobile devices trying to connect to an Umbrel running at home for personal wallet usage via Tor.

Intention is to get RPC calls through Tor via cookie authentication through a Socks5 transport.

When I tried implementing it I ended up recreating the same `SimpleHttpTransport` for Socks5 in this PR in BDK
https://github.com/bitcoindevkit/bdk/pull/493

The resting `Transport` looks like this.  
https://github.com/bitcoindevkit/bdk/blob/b99f1cf6309c3daa6b7d432ecd77bbc69862af44/src/blockchain/rpc_proxy.rs

But its essentially duplicating code from low level, for an high level application.

Opening this issue to start discussing on the best approach to integrate it in `jsonrpc` level and then in `bitcoincore-rpc`.

My suggestion would be to add another module name `simple_socks5` and export that back to `bitcoincore-rpc`. Use that specific transport, if proxy options are set in client config. (Also maybe via a feature flag?)

I am open to work on the PR for this crate and for `bitcoincore-rpc`, if we can decide on the approach.

apoelstra:
This seems like a good idea. I notice that your `socks` dependency looks pretty solid: it has only a couple dependencies, all of which are long-stable and well maintained. One is `byteorder` which broke our MSRV unexpectedly in the past, but this is not a hill worth dying on.

I think the first step would be to update the MSRV of **this** library from 1.29 and 1.41, modernize the code a bit.

@tcharding are you willing to take a look at this? I think we're close to being able to pull this project under the rust-bitcoin org, it just needs a little bit of love.

rajarshimaitra:
Thanks @apoelstra for the look.. I believe it can be a really useful feature not just for BDK but for core rpc in general too.. There is really no good way to use home nodes under dynamic IPs for personal usages without Tor.

Wallet devs mostly need to cook up their own solutions today.. But having native support in bitcoincore-rpc will be very helpful for downstream usage..

tcharding:
> @tcharding are you willing to take a look at this?

Sure thing, no sweat.

tcharding:
Done: https://github.com/apoelstra/rust-jsonrpc/pull/58

A lot of patches but should all be pretty easy to review.

[...]

PR #70, opened 2022-08-29
Add SOCKS5 Proxy support in http transport

rajarshimaitra:
This is an attempt at #57. Which was a request for adding socks5 Proxy support into the http transport. 

This is useful to connect to RPC via Tor.

Summary:
 - Adds a new feature flag `proxy` which does some extra steps to connect the `url` address via a `proxy` address.
 - New fields are added inside `SimpleHttpTransport` which are only visible in `proxy` feature.
 - The first PR contains a refactoring which takes the URL sanity checking logic to its own function. So that it can be used for the `proxy-addr` checking also.
 - New constructor:  Client::http_proxy()
 - Test covering basic behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ simple_http = [ "base64" ]
 simple_tcp = []
 # Basic transport over a raw UnixStream
 simple_uds = []
+# Enable Socks5 Proxy in transport
+proxy = ["socks"]
 
 
 [dependencies]
@@ -30,3 +32,4 @@ serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = [ "raw_value" ] }
 
 base64 = { version = "0.13.0", optional = true }
+socks = { version = "0.3.4", optional = true}
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -3,7 +3,8 @@
 //! if minimal dependencies are a goal and synchronous communication is ok.
 
 use std::io::{BufRead, BufReader, Write};
-use std::net::{TcpStream, ToSocketAddrs};
+use std::net::TcpStream;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::time::{Duration, Instant};
 use std::{error, fmt, io, net, thread};
 
@@ -246,6 +247,65 @@ fn get_lines<R: BufRead>(reader: &mut R) -> Result<String, Error> {
     Ok(body)
 }
 
+/// Do some very basic manual URL parsing because the uri/url crates
+/// all have unicode-normalization as a dependency and that's broken.
+fn check_url(url: &str) -> Result<(SocketAddr, String), Error> {
+    // The fallback port in case no port was provided.
+    // This changes when the http or https scheme was provided.
+    let mut fallback_port = DEFAULT_PORT;
+
+    // We need to get the hostname and the port.
+    // (1) Split scheme
+    let after_scheme = {
+        let mut split = url.splitn(2, "://");
+        let s = split.next().unwrap();
+        match split.next() {
+            None => s, // no scheme present
+            Some(after) => {
+                // Check if the scheme is http or https.
+                if s == "http" {
+                    fallback_port = 80;
+                } else if s == "https" {
+                    fallback_port = 443;
+                } else {
+                    return Err(Error::url(url, "scheme should be http or https"));
+                }
+                after
+            }
+        }
+    };
+    // (2) split off path
+    let (before_path, path) = {
+        if let Some(slash) = after_scheme.find('/') {
+            (&after_scheme[0..slash], &after_scheme[slash..])
+        } else {
+            (after_scheme, "/")
+        }
+    };
+    // (3) split off auth part
+    let after_auth = {
+        let mut split = before_path.splitn(2, '@');
+        let s = split.next().unwrap();
+        split.next().unwrap_or(s)
+    };
+
+    // (4) Parse into socket address.
+    // At this point we either have <host_name> or <host_name_>:<port>
+    // `std::net::ToSocketAddrs` requires `&str` to have <host_name_>:<port> format.
+    let mut addr = match after_auth.to_socket_addrs() {
+        Ok(addr) => addr,
+        Err(_) => {
+            // Invalid socket address. Try to add port.
+            format!("{}:{}", after_auth, fallback_port).to_socket_addrs()?
+        }
+    };
+
+    match addr.next() {
+        Some(a) => Ok((a, path.to_owned())),
+        None => Err(Error::url(url, "invalid hostname: error extracting socket address")),
+    }
+}
+
 impl Transport for SimpleHttpTransport {
     fn send_request(&self, req: Request) -> Result<Response, crate::Error> {
         Ok(self.request(req)?)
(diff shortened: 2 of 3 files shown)


src/simple_http.rs
```diff
@@ -362,6 +362,21 @@ impl Builder {
         self
     }
 
+    /// Add proxy address to the transport for SOCKS5 proxy
+    #[cfg(feature = "proxy")]
```
tcharding, line 399: This stuff might have slipped into the wrong patch? The "proxy" feature is added in the following patch.

src/simple_http.rs
```diff
@@ -282,66 +341,7 @@ impl Builder {
...
-            }
-        };
-
-        self.tp.addr = match addr.next() {
-            Some(a) => a,
-            None => {
-                return Err(Error::url(url, "invalid hostname: error extracting socket address"))
-            }
-        };
-        self.tp.path = path.to_owned();
```
tcharding, line 344: The `self.tp.path` field is not being set in patch 1, so the logic has changed - is this what we wanted?

src/simple_http.rs
```diff
@@ -246,6 +246,65 @@ fn get_lines<R: BufRead>(reader: &mut R) -> Result<String, Error> {
     Ok(body)
 }
 
+/// Do some very basic manual URL parsing because the uri/url crates
+/// all have unicode-normalization as a dependency and that's broken.
+fn check_url(url: &str) -> Result<SocketAddr, Error> {
```
tcharding, line 284: Can we have a unit test please added as a separate patch before the first refactor patch to test `url()` - that way we know that the refactor does not introduce any logic changes.

Cargo.toml
```diff
@@ -23,10 +23,13 @@ simple_http = [ "base64" ]
 simple_tcp = []
 # Basic transport over a raw UnixStream
 simple_uds = []
+# Enable Socks5 Proxy in transport
+proxy = ["socks"]
```
tcharding, line 27: Can we have testing of the "proxy" feature added to ci please (in `contrib/test.sh`).

src/simple_http.rs
```diff
@@ -362,6 +395,21 @@ impl Builder {
...
+    /// Add proxy address to the transport for SOCKS5 proxy
+    #[cfg(feature = "proxy")]
+    pub fn proxy_addr<S: AsRef<str>>(mut self, proxy_addr: S) -> Result<Self, Error> {
+        self.tp.proxy_addr = check_url(proxy_addr.as_ref())?;
+        Ok(self)
+    }
+
+    /// Add optional proxy authentication as ('username', 'password')
+    #[cfg(feature = "proxy")]
+    pub fn proxy_auth<S: AsRef<str>>(mut self, proxy_auth: Option<(S, S)>) -> Self {
```
tcharding, line 407: Its not immediately clear to me what the point of the `Option` is? Is the point so that `proxy_auth(None)` can be used to clear the proxy auth? If so perhaps document that. If that is not the use then perhaps we should use args `(user: S, pass: S)`.

tcharding (commented):
Thanks @rajarshimaitra, looks pretty good, just a few suggestions.

src/simple_http.rs
rajarshimaitra: My bad.. Parallax error.. Fixed..

src/simple_http.rs
rajarshimaitra: Uuuff.. This one was bad.. Almost broke the whole thing.. Thanks, added in https://github.com/apoelstra/rust-jsonrpc/pull/70/commits/bb7b26f4d3879c4f1f316eb3257438dbb9cb8a73

src/simple_http.rs
rajarshimaitra: I was thinking about the same. But these tests here https://github.com/apoelstra/rust-jsonrpc/blob/e3d92e30e1a6f04cbd2278f05a41966f203a65a2/src/simple_http.rs#L400-L451

seemed to be covering the the new moved codes too.. So in theory we are having the same test coverage.. And as there wasn't (suppose to be) any changed logic, so I couldn't figure what new tests to write..
But its interesting that we didn't fail even when `self.tp.path` wasn't being set properly. Maybe that's the compiler telling us we need more tests?

Cargo.toml
rajarshimaitra: Should have seen those one myself.. Done..

src/simple_http.rs
rajarshimaitra: You are right. Having `Option<>` doesn't make much sense here.. Updated to your second suggestion..

src/simple_http.rs
rajarshimaitra: I added some basic assertions that we are setting the values correctly here.. https://github.com/apoelstra/rust-jsonrpc/pull/70/commits/9f10f71a2f51e32953d52d1265522b5a61b37e78

tcharding:
We are already abusing the test environment in `run.sh` (killing bitcoind processes etc.), you could continue on that theme and spin up a bitcoind process behind tor proxy then connect to it. I did something like this before when writing a libp2p tor transport and I don't remember having any trouble.

rajarshimaitra:
> We are already abusing the test environment in run.sh (killing bitcoind processes etc.), you could continue on that theme and spin up a bitcoind process behind tor proxy then connect to it.

I didn't fully understood this part. Which `run.sh` we are talking about? 

Also I pushed a trial patch in https://github.com/apoelstra/rust-jsonrpc/pull/70/commits/8e31ceaae03a48d439e894c8014161dcde0248e2 to connect with a real bitcoind node using the `electrsd` crate.. But I am observing this failure.. Any suggestion on this? 

```
running 1 test
thread 'simple_http::tests::test_connection' panicked at 'called `Result::unwrap()` on an `Err` value: Transport(SocketError(Custom { kind: Other, error: "general SOCKS server failure" }))', src/simple_http.rs:578:60
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/result.rs:1805:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/result.rs:1098:23
   4: jsonrpc::simple_http::tests::test_connection
             at ./src/simple_http.rs:578:24
   5: jsonrpc::simple_http::tests::test_connection::{{closure}}
             at ./src/simple_http.rs:569:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:248:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test simple_http::tests::test_connection ... FAILED
```

rajarshimaitra:
Thanks @tcharding for the suggestion. I didn't fully understand the suggestion. Which `run.sh` we are talking about here..

I tried another approach with the `electrsd` crate and connected the client to a bitcoind node, in this trial patch here https://github.com/apoelstra/rust-jsonrpc/pull/70/commits/c367e757656583439e513ae5f5db5390d155ec84. Were you suggesting something similar??

tcharding:
> Which run.sh we are talking about here..

Woops sorry I got mixed up with `rust-bitocincore-rpc` (which has `integration_test/run.sh`).

In general I do not like unit tests modifying the environment they run in. The ways out of this, none of which are that nice IMO, are to use docker (this makes running tests a bit of a PITA though) and/or put all the environment changing stuff in a bash script (like we do in `rust-bitcoincore-rpc` at least then folks should read it before running it and see it mangles their environment. Perhaps @apoelstra has preferences or ideas other than this, please do not just implement my opinions on this :)

rajarshimaitra:
Thanks @tcharding for the explanation. I was also thinking, we need to expose this ultimately through `rust-bitcoincore-rpc`. So maybe better to handle the test over there? And this changes can go with the unit tests only in this crate?

apoelstra:
Yeah, it may be simplest to just let rust-bitcoincore-rpc do the testing for us. This would lead to a delay in catching bugs but it's a lot better than nothing.

I wouldn't mind a script that detected if it was running inside cirrus and then spawned a docker container. But I definitely wouldn't run docker stuff locally, even if I had the spare RAM to run zombie instances indefinitely I don't trust the "download an entire operating system and run it on your hardware" model :)

rajarshimaitra:
Thanks @apoelstra for the suggestion.. I would start working on exposing this via rust-bitcoincore-rpc.. I reverted the `bitcoind` test instance and applied all the suggestion by @tcharding .. So does this PR looks ready on its own?

tcharding:
Can you squash this down so that the patch set does not show the development process of the PR please.

apoelstra:
Done reviewing 9f10f71a2f51e32953d52d1265522b5a61b37e78. I don't mind the commit structure as-is, but I do need every commit to compile and pass unit tests, which is currently not the case. Otherwise ACK from me.

rajarshimaitra:
Thanks @tcharding and @apoelstra for the reviews.. I have updated the commit history and consolidated in two parts.. I think the compilation and tests are now passing at each commit..

apoelstra:
@rajarshimaitra sadly, your assignment to a tuple does not compile with Rustc 1.41, which is our MSRV. Can you split that line up into multiple assignments?

rajarshimaitra:
Ah my mistake.. I forgot to check it against MSRV.. Updated in the last push..

apoelstra (approved):
ACK ce8f318a9984b921c98a687bc33d434fdd36a80c

Issue #67, opened 2022-08-05
Timeout on MacOS after making ~16384 RPC calls


casey:
I originally hit this while using `rust-bitcoincore-rpc`, but was able to narrow it down to this crate.

The error is:

```
Transport(SocketError(Error { kind: TimedOut, message: "connection timed out" }))
```

I uploaded a repo containing a self-contained reproduction here: https://github.com/casey/bitcoind-rpc-timeout/blob/master/src/main.rs

The issue was introduced some time between 0.11 and 0.12 and is still present on master. I tried to bisect it down, but there was a lot of flux in the crate, which made it impossible to track down a specific revision.

While debugging, I would sometimes get this error:

```
Transport(SocketError(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }))
```

Which makes me think it's related to this issue: https://github.com/rust-bitcoin/rust-bitcoincore-rpc/issues/211
And this issue: https://github.com/bitcoin/bitcoin/issues/25724#issuecomment-1198408118

Although I wouldn't necessarily bet on it.

tcharding:
That repo crate should hit the bug when running using `cargo run`, right? Just posting here to confirm the `MacOS` in the title, on Ubuntu 20.04 the bug is not triggered.

casey:
> That repo crate should hit the bug when running using `cargo run`, right? Just posting here to confirm the `MacOS` in the title, on Ubuntu 20.04 the bug does not appear to be triggered.

Yup, that's right. I also tried it on an Ubuntu 21.10 server, and wasn't able to reproduce it there.

casey:
For reference, I added a GitHub actions workflow that [passes on Ubuntu](https://github.com/casey/bitcoind-rpc-timeout/runs/7684043094?check_suite_focus=true) and [fails on MacOS](https://github.com/casey/bitcoind-rpc-timeout/runs/7684043206?check_suite_focus=true).

tcharding:
Good bug reporting effort bro, thanks!

apoelstra:
Between 0.11 and 0.12 we completely replaced hyper with a homebrew http transport. Probably we are leaking some resource.

apoelstra:
Yep -- those are my guesses. I'll try to poke at this later today, although I don't have a Mac so it might be a bit annoying.

On your Mac could you try running your test in `strace` and see if there's some particular syscall (like `open` on a socket?) which is timing out?

casey:
[...]

PR #72, opened 2022-10-26
Reuse socket

raphjaph:
This is a quick and dirty attempt at fixing #67 that reuses the socket for multiple requests. It has few issues: 
- no error handing, client becomes unusable after encountering a single error
- `SimpleHttpTransport::request()` takes a `mut` reference to self, which cascades to all downstream users
- `TcpStream` is not `Clone`, therefore I had to remove the derive `Clone`

An alternate implementation based on `reqwest` (`hyper` seems too low-level because of the async stuff) wouldn't have these problems.  So I think I'm going to start on that. Just wanted so share this draft PR if anyone has feedback or has an idea on how to do this without `reqwest`.

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -4,10 +4,10 @@
 
 #[cfg(feature = "proxy")]
 use socks::Socks5Stream;
-use std::io::{BufRead, BufReader, Write};
-#[cfg(not(feature = "proxy"))]
+use std::io::{BufRead, BufReader, Read, Write};
 use std::net::TcpStream;
 use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use std::{error, fmt, io, net, thread};
 
@@ -38,6 +38,7 @@ pub struct SimpleHttpTransport {
     proxy_addr: net::SocketAddr,
     #[cfg(feature = "proxy")]
     proxy_auth: Option<(String, String)>,
+    sock: Arc<Mutex<Option<TcpStream>>>,
 }
 
 impl Default for SimpleHttpTransport {
@@ -57,6 +58,7 @@ impl Default for SimpleHttpTransport {
             ),
             #[cfg(feature = "proxy")]
             proxy_auth: None,
+            sock: Arc::new(Mutex::new(None)),
         }
     }
 }
@@ -76,26 +78,55 @@ impl SimpleHttpTransport {
     where
         R: for<'a> serde::de::Deserialize<'a>,
     {
+        // `try_request` should not panic, so the mutex shouldn't be poisoned
+        // and unwrapping should be safe
+        let mut sock = self.sock.lock().expect("poisoned mutex");
+        match self.try_request(req, &mut sock) {
+            Ok(response) => Ok(response),
+            Err(err) => {
+                *sock = None;
+                Err(err)
+            }
+        }
+    }
+
+    fn try_request<R>(
+        &self,
+        req: impl serde::Serialize,
+        sock: &mut Option<TcpStream>,
+    ) -> Result<R, Error>
+    where
+        R: for<'a> serde::de::Deserialize<'a>,
+    {
         // Open connection
         let request_deadline = Instant::now() + self.timeout;
-        #[cfg(feature = "proxy")]
-        let mut sock = if let Some((username, password)) = &self.proxy_auth {
-            Socks5Stream::connect_with_password(
-                self.proxy_addr,
-                self.addr,
-                username.as_str(),
-                password.as_str(),
-            )?
-            .into_inner()
-        } else {
-            Socks5Stream::connect(self.proxy_addr, self.addr)?.into_inner()
+        if sock.is_none() {
+            *sock = Some({
+                #[cfg(feature = "proxy")]
+                {
+                    if let Some((username, password)) = &self.proxy_auth {
+                        Socks5Stream::connect_with_password(
+                            self.proxy_addr,
+                            self.addr,
+                            username.as_str(),
+                            password.as_str(),
+                        )?
+                        .into_inner()
+                    } else {
+                        Socks5Stream::connect(self.proxy_addr, self.addr)?.into_inner()
+                    }
+                }
+
+                #[cfg(not(feature = "proxy"))]
+                {
+                    let stream = TcpStream::connect_timeout(&self.addr, self.timeout)?;
+                    stream.set_read_timeout(Some(self.timeout))?;
+                    stream.set_write_timeout(Some(self.timeout))?;
+                    stream
+                }
+            })
         };
-
-        #[cfg(not(feature = "proxy"))]
-        let mut sock = TcpStream::connect_timeout(&self.addr, self.timeout)?;
-
-        sock.set_read_timeout(Some(self.timeout))?;
-        sock.set_write_timeout(Some(self.timeout))?;
+        let sock = sock.as_mut().unwrap();
 
         // Serialize the body first so we can set the Content-Length header.
         let body = serde_json::to_vec(&req)?;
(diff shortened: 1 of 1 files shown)


apoelstra:
Cool! Thanks for taking this on.

Honestly I think you're close to having a "good" design. A couple small changes would fix your limitations:
* To get error handling, catch errors and set `self.sock` to `None` before returning; now the client can be reused. A later iteration should maybe detect `EINTR` and retry sending automatically, or something, but we can deal with that later.
* To avoid `&mut` and get `Clone` we **could** wrap the underlying socket in a `Arc<Mutex>`...but OTOH the user can do that themselves, which is a minor annoyance and gives the user more freedom.

raphjaph:
Thanks for the tips! Have a look and let me know what you think.

apoelstra:
Could you squash all the commits into one?

src/simple_http.rs
```diff
@@ -73,29 +75,50 @@ impl SimpleHttpTransport {
     }
 
     fn request<R>(&self, req: impl serde::Serialize) -> Result<R, Error>
+    where
+        R: for<'a> serde::de::Deserialize<'a>,
+    {
+        match self.try_request(req) {
+            Ok(response) => Ok(response),
+            Err(err) => {
+                *self.sock.lock().unwrap() = None;
```
apoelstra, line 84: I think we should add a mutex poisoning error to `Error` and then use `?` rather than unwrapping. In theory the user might want to catch this and recover somehow.

src/simple_http.rs
```diff
@@ -73,29 +75,50 @@ impl SimpleHttpTransport {
...
-        sock.set_read_timeout(Some(self.timeout))?;
-        sock.set_write_timeout(Some(self.timeout))?;
+            #[cfg(not(feature = "proxy"))]
+            {
+                let stream = TcpStream::connect_timeout(&self.addr, Duration::from_secs(15))?;
+                stream.set_read_timeout(Some(self.timeout))?;
+                stream.set_write_timeout(Some(self.timeout))?;
+                *sock = Some(stream);
+            }
+        }
```
apoelstra, line 120: Could you refactor this code to instead look like
```
let stream =
#[cfg(feature = "proxy")] {
    // ...
};
#[cfg(not(feature = "proxy"))] {
    // ...
};
*sock = Some(stream);
```
So that both cases have the same side-effect profile and it's clear that in both cases you're constructing a new object and assigning it to `*sock`.

src/simple_http.rs
```diff
@@ -73,29 +75,50 @@ impl SimpleHttpTransport {
...
+        match self.try_request(req) {
+            Ok(response) => Ok(response),
+            Err(err) => {
+                *self.sock.lock().unwrap() = None;
+                Err(err)
+            }
+        }
+    }
+
+    fn try_request<R>(&self, req: impl serde::Serialize) -> Result<R, Error>
```
apoelstra, line 90: We should wrap this in another function which intercepts every error and assigns `*sock = None` before returning.

src/simple_http.rs
raphjaph: The mutex can only be poisoned if `try_request` panics. I think since we are doing proper error handling in that function it should never happen. But I'm not 100% sure on that.

src/simple_http.rs
raphjaph: I believe using cfg with expressions is experimental so I can't do it that nicely yet. I have made it look a bit cleaner in the updated commit.

src/simple_http.rs
raphjaph: I think we're already doing this; `request()` wraps `try_request()`.

src/simple_http.rs
apoelstra: I considered this, and I'm not 100% either. If you like you can keep the `.unwrap` and just add a comment saying that we're pretty confident that we'll never panic when the mutex is open.

src/simple_http.rs
apoelstra: Oh, derp, yes

apoelstra (approved):
ACK 9f4ab54bcc8e596bb9644f78590826a6ab92b7d3

raphjaph:
We tried using this and something went wrong, so definitely don't merge this yet. I'll try to take a look at this over the weekend and figure out what exactly went wrong.

raphjaph:
There were two problems:
1. The `Connection: Close` HTTP Header was set.  This would tell the RPC server to close the connection after it finished writing, so subsequent socket writes would beget `broken pipe` errors. 
2. We needed to parse the `Content-Length` Header to know when the response was done. 

We tested it with our application and everything works as it should now. So this would also close #67.

src/simple_http.rs
```diff
@@ -73,29 +75,58 @@ impl SimpleHttpTransport {
     }
 
     fn request<R>(&self, req: impl serde::Serialize) -> Result<R, Error>
+    where
+        R: for<'a> serde::de::Deserialize<'a>,
+    {
+        // `try_request` should not panic, so the mutex shouldn't be poisoned
+        // and unwrapping should be safe
+        let mut sock = self.sock.lock().unwrap();
```
tcharding, line 83:
```suggestion
        let mut sock = self.sock.lock().expect("poisoned mutex");
```

tcharding (approved):
ACK c621b972ecb20ddfee1f611c7cfb4c6ba3c376b9

tcharding:
ACK 1826d5e7ec02d9dd3fbca94bfa3b4215870d193f

tcharding:
Gentle bump @apoelstra

src/simple_http.rs
```diff
@@ -133,18 +163,39 @@ impl SimpleHttpTransport {
...
+        }
 
         if response_code == 401 {
             // There is no body in a 401 response, so don't try to read it
             return Err(Error::HttpErrorCode(response_code));
         }
 
+        let content_length = content_length.ok_or(Error::HttpParseError)?;
+
+        let mut buffer = vec![0; content_length];
```
apoelstra, line 193: We definitely need to put a limit on this, but it's fine for now :). We'll do a followup PR to audit this because I think `get_line` may also be subject to memory exhaustion attacks.

apoelstra (approved):
ACK 1826d5e7ec02d9dd3fbca94bfa3b4215870d193f

apoelstra:
Hmm, with this PR I see a dramatic (~10x) slowdown in my syncing speed. `top` shows around 20% CPU usage from each of `icboc` (my wallet) and `bitcoind`, i.e. both are only using one core. Vs with the commit before this PR both are close to 50%.

My ACK still stands because we need to fix the Mac bug but we need to investigate this.

apoelstra:
Ok, by tightening the time we have the socket locked, the perf hit is only 50-100% rather than 10x and it seems like my CPU is pretty close to being maxed out again. I'm fine with this, since really improving performance will require we maintain a pool of sockets, which is a whole nother thing.

I'm running these numbers on an unplugged laptop so 50-100% may well be "no change" :)

tcharding:
@raphjaph just fyi your CI fails are not your fault, will be resolved by: https://github.com/apoelstra/rust-jsonrpc/pull/74#issuecomment-1319163806

apoelstra:
@raphjaph can you rebase on master? We've fixed the CI.

src/simple_http.rs
raphjaph: yes, that sounds prudent

raphjaph:
> _Edit: a previous version of this comment said that there was still a 50-100% perf hit. This is because my laptop was unplugged and therefore throttled when I was doing the measurement._
> 
> Ok, after modifying this PR to tighten the mutex lock's scope, in particular to unlock it before doing any parsing of the result, I get the same performance as before (possibly a small speedup, but I'm not measuring carefully enough to tell). I'm fine to move that to a followup PR as well.

Ok, that's good to know I'll do the performance stuff in a follow-up PR then.

apoelstra (approved):
ACK e86737473fbdbafa49c4a1e4d3e6657a8ed42446

PR #74, opened 2022-11-16
Clear clippy warnings

tcharding:
Clear all the clippy warnings introduce by a recent clippy update.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,14 +6,15 @@ jobs:
   Tests:
     name: Tests
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
         toolchain:
-          - 1.41.1
           - stable
+          - 1.41.1
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout Crate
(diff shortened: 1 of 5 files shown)


tcharding:
Rebased on #75, seems to work.

.github/workflows/rust.yml
```diff
@@ -27,7 +28,7 @@ jobs:
       - name: Running tests on ${{ matrix.toolchain }}
```
casey, line 28: You can do this to avoid running on macos:

```
if: ${{ matrix.os != 'macOS-latest' }}
```

Also, I've seen it spelled `macos-latest` before, but maybe that doesn't matter.

apoelstra: How do I prevent specifically rustc 1.41 from running on macos?

apoelstra (approved):
ACK 4fd4cb10eaa4243e85586e5e3ac36315c02849c1

.github/workflows/rust.yml
casey: Ah, sorry, I didn't realize that the combination of  macos and 1.41. I didn't test this, but it should be something like:

```
if: ${{ !(matrix.os == 'macOS-latest' && matrix.toolchain == `1.4.1`) }}
```

apoelstra: It's a huge PITA to test github actions :) bash is easier.

PR #75, opened 2022-11-17
WIP: try to fix CI on macos

apoelstra:
It looks like there is some old LLVM/rustc/linker/??? bug that is preventing `syn` from building on MacOS and breaking our CI. This commit just disables it.

cc @tcharding can you pull this commit into #74 and then we'll merge that?

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -11,6 +11,15 @@ if cargo --version | grep nightly; then
     NIGHTLY=true
 fi
 
+# On MacOS on 1.41 we get a link failure with syn
+# This is fixed by https://github.com/rust-lang/rust/pull/91604 (I think)
+# but implies that we can't do testing on MacOS for now, at least with 1.41.
+if cargo --version | grep "1\.41\.0"; then
+    if [ "$RUNNER_OS" = "macOS" ]; then
+        exit 0
+    fi
+fi
+
 # Defaults / sanity checks
 cargo build --all
 cargo test --all
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,14 +6,15 @@ jobs:
   Tests:
     name: Tests
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
         toolchain:
-          - 1.41.1
           - stable
+          - 1.41.1
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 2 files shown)


apoelstra:
Results so far:

* Previously we were just running `cargo test --all-features` instead of the full feature matrix, for some reason. I fixed that, though it seems to have had no effect, for good or bad.
* The failure doesn't happen on MacOS with the latest rustc, nor does it happen on Windows or Ubuntu
* It happens with `syn` as far back as 1.0.90 (the earliest version that `serde` will let me use). This came out 8 months ago, so it seems like something has broken with CI rather than something breaking with `syn`

contrib/test.sh
```diff
@@ -11,6 +11,15 @@ if cargo --version | grep nightly; then
     NIGHTLY=true
 fi
 
+# On MacOS on 1.41 we get a link failure with syn
+# This is fixed by https://github.com/rust-lang/rust/pull/91604 (I think)
+# but implies that we can't do testing on MacOS for now, at least with 1.41.
+if cargo --version | grep "1\.41\.0"; then
```
tcharding, line 17: This seems to be working but we are running with toolchain 1.41.1, why is the regex statement succeeding?

tcharding: This shows it works: `Continuous integration / Tests (macOS-latest, 1.41.1) (pull_request) Successful in 15s `

tcharding: Bizzare, this is the version of cargo used in the CI pipeline: cargo 1.41.0 (626f0f40e 2019-12-03)

apoelstra: It "works" because we `exit 0` here.

tcharding: lolz, I meant why does `cargo --version` report "1\.41\.0" when we are using the 1.41.1 toolchain.

tcharding: Said another way, the regex surprised me because of the zero in the toolchain version.

apoelstra: Oh, hmm, I wonder if this `grep` is doing the opposite of what I want (and skipping every test)

apoelstra: Oops, I forgot to reply to this... the reason is that **cargo** is at 1.41.0 even though rustc is at 1.41.1.

tcharding: oooh, I see. Thanks

PR #76, opened 2022-11-19
Simple HTTP cleanups and followup to #72

apoelstra:
Tightens the mutex lifetime to improve performance. Basically we read all the data into a buffer and then unlock the mutex ASAP. We could be a bit more memory-efficient by using `serde_json::from_reader` to parse directly from the socket, but (a) that would make it harder to enforce the Content-length header; and (b) it'd hold the mutex for longer than necessary.

This commit also splits out the `HttpParseError` into several variants, which unfortunately makes this a breaking change. I think I can move that commit into its own PR if we want to get a minor rev out, but I don't think we care too much about that since this crate is pretty far to the edge of most dependency trees.

Adds a fuzz harness but doesn't integrate it into CI. Fixes a couple bugs I found while fuzzing locally.

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -9,7 +9,7 @@ use std::io::{BufRead, BufReader, BufWriter, Read, Write};
 #[cfg(not(fuzzing))]
 use std::net::TcpStream;
 use std::net::{SocketAddr, ToSocketAddrs};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::{Duration, Instant};
 use std::{error, fmt, io, net, num, thread};
 
@@ -145,9 +145,9 @@ impl SimpleHttpTransport {
         // Open connection
         let request_deadline = Instant::now() + self.timeout;
         // No part of this codebase should panic, so unwrapping a mutex lock is fine
-        let mut sock = self.sock.lock().expect("poisoned mutex");
-        if sock.is_none() {
-            *sock = Some(BufReader::new({
+        let mut sock_lock: MutexGuard<Option<_>> = self.sock.lock().expect("poisoned mutex");
+        if sock_lock.is_none() {
+            *sock_lock = Some(BufReader::new({
                 #[cfg(feature = "proxy")]
                 {
                     if let Some((username, password)) = &self.proxy_auth {
@@ -174,7 +174,7 @@ impl SimpleHttpTransport {
         };
         // In the immediately preceding block, we made sure that `sock` is non-`None`,
         // so unwrapping here is fine.
-        let mut sock = sock.as_mut().unwrap();
+        let mut sock: &mut BufReader<_> = sock_lock.as_mut().unwrap();
 
         // Serialize the body first so we can set the Content-Length header.
         let body = serde_json::to_vec(&req)?;
@@ -254,8 +254,8 @@ impl SimpleHttpTransport {
         let buffer = match content_length {
             None => {
                 let mut buffer = Vec::with_capacity(INITIAL_RESP_ALLOC as usize);
-                // `take` consumes `sock` and drops it, unlocking the mutex
                 sock.take(FINAL_RESP_ALLOC).read_to_end(&mut buffer)?;
+                drop(sock_lock);
                 buffer
             },
             Some(n) if n > FINAL_RESP_ALLOC => {
@@ -266,8 +266,8 @@ impl SimpleHttpTransport {
             },
             Some(n) => {
                 let mut buffer = Vec::with_capacity(INITIAL_RESP_ALLOC as usize);
-                // `take` consumes `sock` and drops it, unlocking the mutex
                 let n_read = sock.take(n).read_to_end(&mut buffer)? as u64;
+                drop(sock_lock);
                 if n_read < n {
                     return Err(Error::IncompleteResponse { content_length: n, n_read });
                 }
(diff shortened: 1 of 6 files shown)


apoelstra:
cc @raphjaph does this do what you want?
cc @tcharding for review

apoelstra:
This PR does not recover the pre-#72 performance. I'm investigating.

apoelstra:
Honestly I'm not sure what to make of this. It looks like for some reason Bitcoin Core takes ~10ms to reply to `getblock` calls prior to the socket reuse patch, and ~60ms after. This is visible with `debug=1` and `logtimemicros=1` on a Core node, e.g. as
```
2022-11-19T19:31:06.854923Z [rpc] ThreadRPCServer method=getblock user=__cookie__
2022-11-19T19:31:06.895467Z [libevent:debug] evhttp_add_header: key: Content-Type val: application/json
```
where that `add_header` is the header of its own response.

It has nothing to do with the mutex on our end, or any sort of buffering, as near I can tell. I have tried it with the tightened mutex scope, with socket pools of various sizes, and it seems like it's quick to reply to the first RPC call on a connection, and slow to reply to every one after that. Although there is some noise so it's hard to be certain.

apoelstra:
Ok, I can get most of the performance back by both:
* Storing the `BufReader` rather than creating a new one on every request. (This is actually a bug, since when doing so, we potentially read further extra data from the stream that we simply drop.)
* Calling `serde_json::from_reader` rather than reading the data into a buffer then calling `serde_json::from_slice`

I don't know why the latter is necessary when it wasn't under the old "create a new connection every time" model, and I don't know why I need to make both changes when only the latter should provide any visible change under normal circumstances.

src/simple_http.rs
```diff
@@ -155,31 +202,43 @@ impl SimpleHttpTransport {
         let mut reader = BufReader::new(sock);
 
         // Parse first HTTP response header line
-        let http_response = get_line(&mut reader, request_deadline)?;
-        if http_response.len() < 12 || !http_response.starts_with("HTTP/1.1 ") {
-            return Err(Error::HttpParseError);
+        let http_response = get_line(&mut reader, Instant::now() + self.timeout)?;
```
tcharding, line 205: What's the reason not to use `request_deadline` here? If there is one can be have a code comment please?

src/simple_http.rs
```diff
@@ -189,16 +248,40 @@ impl SimpleHttpTransport {
...
+                buffer
+            },
+            Some(n) if n > FINAL_RESP_ALLOC => {
+                return Err(Error::HttpResponseContentLengthTooLarge {
+                    length: n,
+                    max: FINAL_RESP_ALLOC,
+                });
+            },
+            Some(n) => {
+                let mut buffer = Vec::with_capacity(INITIAL_RESP_ALLOC as usize);
```
tcharding, line 268: Shouldn't the buffer have capacity `n`?

tcharding (commented):
Couple minor questions but apart from that looks good to me.

src/simple_http.rs
apoelstra: Yeah, I think so. I went back and forth a couple times on this, but now that we bound `n` I think we should use it.

src/simple_http.rs
apoelstra: I think I meant to change `request_deadline` everywhere, and it's a rebasing mistake that it hasn't been changed.

tcharding: :)

src/simple_http.rs
```diff
@@ -19,13 +20,55 @@ use serde_json;
...
+/// Global mutex used by the fuzzing harness to inject data into the read
+/// end of the TCP stream.
+pub static FUZZ_TCP_SOCK: Mutex<Option<io::Cursor<Vec<u8>>>> = Mutex::new(None);
+
+#[cfg(fuzzing)]
+#[derive(Clone, Debug)]
+struct TcpStream;
+
+#[cfg(fuzzing)]
+mod impls {
```
raphjaph, line 33: What is the reason to put the implementation of the fuzzed TcpStream into its own module?

I'm new to Rust so forgive me if this is a silly question.

raphjaph:
Looks good to me. 

I don't 100% understand how the fuzzing works but have never seen that so it was cool to see that in the wild.

src/simple_http.rs
tcharding: So that the `#[cfg(fuzzing)]` is applied to the whole module.

raphjaph: thanks, that makes sense. 

The `TcpStream` is  declared outside of the `impls` mod so it doesn't have have to be imported explicitly in the `simple_http` mod.

apoelstra:
Thanks @raphjaph !

Unfortunately I am going to rebase this and change a few things (hopefully not a huge code diff). So I may need you to review again.

apoelstra:
@tcharding I have pushed 4 new commits which address your nits and get us some further perf improvements.

Weirdly this is still not as fast as the old "new connection for every request" logic, even though it's *way* more intelligent about buffering, but we're within a factor 2 now, and we're able to completely pin my CPU (the old code was 50/50 Core vs icboc, this code is 30/70, so the implication is that we're doing more processing and that's why we're slower ... but why?) so I'm okay with it.

apoelstra:
I think the performance hit is caused by https://github.com/serde-rs/json/issues/160#issuecomment-253446892 ... but I don't have a good fix because reading from the socket into memory prior to starting parsing is actually slower in this case.

I think we should just leave this be until somebody complains, then maybe we need to write our own json deserializer or something.

tcharding (approved):
ACK 2380d90c8f8d4795805de199f23a8cc70fa6d13b

raphjaph:
ACK 2380d90c8f8d4795805de199f23a8cc70fa6d13b

Issue #79, opened 2022-12-07
HttpResponseTooShort for first request after a while


jorian:
I have this strange behaviour with a Discord bot that I'm currently making.

For a cryptocurrency project, I am using rust-jsonrpc as a client to query that cryptocurrency's blockchain. It uses the same RPC mechanism that Bitcoin uses, so I'm using this crate and it has been working fine for quite some time.

Because I have to do many RPCs, I store an instance of a rust-jsonrpc client as global data of the bot. This global data is essentially a `Box::pin(async move { Data } )` where I store the client as a member of Data. (the bot is made using [poise](https://docs.rs/poise/latest/poise/#quickstart), and in [this](https://docs.rs/poise/latest/src/poise/framework/builder.rs.html#68-81) `setup` function is where Data gets initiated)

When I start the bot and keep it running for a while (30+ sec) and get the client from global data and do a `getnewaddress` RPC, I am getting a HttpResponseTooShort error: `error: JsonRPC(Transport(HttpResponseTooShort { actual: 0, needed: 12 }))`. It happens every time I use this call after at least 30 seconds of idling and I can't figure out why this happens. A second request within 30 seconds works just fine.

I've already checked if the problem is keeping a client around for a while:

```rs
    let client = vrsc_rpc::Client::vrsc(true, Auth::ConfigFile).unwrap();

    thread::sleep(Duration::from_secs(50));

    dbg!(client.get_new_address().unwrap());
```
_(I built a wrapper like https://github.com/rust-bitcoin/rust-bitcoincore-rpc for this cryptocurrency project)_

This doesn't appear to be a problem. I can't figure out what is, though. My guess would be that I keep a connection open for too long, but I can't figure this out, as it appears to be working in a separate environment.

Any ideas?

dpc:
Sounds like a connection reuse issue after timeout.

#80 possibly related?

apoelstra:
I agree, it sounds like we're failing to detect that the socket has died. When we do the `POST` request we should be detecting errors (on the second write) and retrying.

From http://www.softlab.ntua.gr/facilities/documentation/unix/unix-socket-faq/unix-socket-faq-2.html
> "If the peer calls close() or exits...I would expect EPIPE, not on the next call, but the one after."

jorian:
I'm glad it looks like the issue is known. What would be the timeline for a fix?

jorian:
What is the status on this?

apoelstra:
[...]

PR #80, opened 2022-12-07
simple_http: actually reuse sockets

darosior:
serde_json::from_reader requires that the other side close the socket for it to complete. Therefore we were forcing the server to always close streams to us, never reusing any single sockets. In fact, it was more a burden because after the first request for every request we would have to call the request() function twice: once for it to notice that the socket was closed on the other side (receiving an EOF on line 202 and confusingly returning an error telling the HTTP header was malformed).

Instead, get back to what was previously used in previous versions: read a line from the stream (using '\n' to mark the end of the request instead of a connection close) and parse it afterward.

From https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html:
> It is expected that the input stream ends after the deserialized object. If the stream does not end, such as in the case of a persistent socket connection, this function will not return. It is possible instead to deserialize from a prefix of an input stream without looking for EOF by managing your own [Deserializer](https://docs.rs/serde_json/latest/serde_json/struct.Deserializer.html).

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -263,7 +263,9 @@ impl SimpleHttpTransport {
         // Attempt to parse the response. Don't check the HTTP error code until
         // after parsing, since Bitcoin Core will often return a descriptive JSON
         // error structure which is more useful than the error code.
-        match serde_json::from_reader(&mut reader) {
+        let mut response_buf = String::new();
+        reader.read_line(&mut response_buf)?;
+        match serde_json::from_str(&response_buf) {
             Ok(s) => {
                 if content_length.is_some() {
                     reader.bytes().count(); // consume any trailing bytes


darosior:
Maybe the fact we now have to make every request twice is (part of) the reason for the performance hit mentioned in https://github.com/apoelstra/rust-jsonrpc/pull/76#issuecomment-1328066949?

At least it can't be that we never actually reusing sockets, since we weren't able to in the first place.

apoelstra:
@darosior yes, `from_reader` expects the stream of bytes to end. this is why we call `take` on the socket. There is no requirement that the other side closes the socket.

apoelstra:
This PR reverts #69.

darosior:
Closing for now. I don't remember the context anymore, but i think #79 is related.

PR #84, opened 2023-01-25
simple_http: when writing fails, try (once) to reconnect the socket

### LINKED ISSUE #79: HttpResponseTooShort for first request after a while

apoelstra:
Attempt at #79. Does not fix the issue, but will merge this anyway since it fixes an issue I've hit in other projects.

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -178,7 +178,18 @@ impl SimpleHttpTransport {
         // Send HTTP request
         {
             let mut write_sock = BufWriter::new(sock.get_mut());
-            write_sock.write_all(b"POST ")?;
+            // When we write to a socket, it may have died but we do not detect it. In this case we
+            // want to detect this ASAP and reconnect. We do this by writing the literal text POST
+            // in two pieces and checking for error returns on either one, and retrying in this
+            // case.
+            //
+            // From http://www.softlab.ntua.gr/facilities/documentation/unix/unix-socket-faq/unix-socket-faq-2.html
+            // "If the peer calls close() or exits...I would expect EPIPE, not on the next call,
+            // but the one after."
+            if write_sock.write_all(b"PO").is_err() || write_sock.write_all(b"ST ").is_err() {
+                **write_sock.get_mut() = self.fresh_socket()?;
+                write_sock.write_all(b"POST ")?;
+            }
             write_sock.write_all(self.path.as_bytes())?;
             write_sock.write_all(b" HTTP/1.1\r\n")?;
             // Write headers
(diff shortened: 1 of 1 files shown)


src/simple_http.rs
```diff
@@ -177,24 +177,36 @@ impl SimpleHttpTransport {
...
+            // case.
+            //
+            // From http://www.softlab.ntua.gr/facilities/documentation/unix/unix-socket-faq/unix-socket-faq-2.html
+            // "If the peer calls close() or exits...I would expect EPIPE, not on the next call,
+            // but the one after."
+            if write_sock.write_all(b"PO").is_err() || write_sock.write_all(b"ST ").is_err() {
+                **write_sock.get_mut() = self.fresh_socket()?;
+                write_sock.write_all(b"POST ")?;
+            }
+            write_sock.write_all(b"POST ")?;
```
tcharding, line 193: Why the second write of "POST" (excluding the error path)? At this stage we've either written POST already or returned an error?

apoelstra: Oh, derp, this second POST should not be there and probably completely breaks things. Pushed to remove it.

tcharding (approved):
ACK 5b7b23c0b3411f758ba2c4249908fbc8208b9e76

PR #85, opened 2023-02-05
Add `host` header to HTTP/1.1

ogenev:
The `host` header for HTTP1.1 must be sent according to [mozilla specs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) and [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority).

The absence of a host header breaks the compatibility with [jsonrpsee](https://github.com/paritytech/jsonrpsee) HTTP server for example.


apoelstra:
Thanks!

apoelstra (approved):
ACK 94aa2090263e79cc4c3192ebe0d57a9049686757

PR #87, opened 2023-03-22
simple_http: improve `request_to_closed_socket` test 

philipr-za:
Make the `request_to_closed_socket` test more robust to parallel execution but having the OS assign a unique port on each run.

It was found if this test was executed in parallel that different instances would connect to the server from an earlier instance.


apoelstra (approved):
ACK 20697ec4f686b76bb692e5bda0c34dc6ecf4e219

PR #90, opened 2023-04-14
simple_http:  handle sockets returning "connection reset on peer" errors

philipr-za:
It was observed via CI that on Mac OS a broken socket could sometimes return the error "Connection reset by peer" on the attempt to read or write when the socket was closed. This behavior was not observed on the *nix and Windows CI jobs but it is a valid behavior. Currently the socket is detected as broken when no bytes are returned by the `read` call.

This PR adds in logic handling the case that when the first write and read operations happen if there was an error then a fresh socket is requested and the write and follow up read is attempted a second time. If the second write and read errors then the error is returned.

### DIFF
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -211,15 +211,23 @@ impl SimpleHttpTransport {
         request_bytes.write_all(&body)?;
 
         // Send HTTP request
-        sock.get_mut().write_all(request_bytes.as_slice())?;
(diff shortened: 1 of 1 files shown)


apoelstra:
Nice!, thanks.

apoelstra (approved):
ACK 11c32a9512316f5c9a405142dfeee7f74ecffbe6

PR #91, opened 2023-05-03
Bump MSRV

tcharding:
- Patch 1 is prep, run the formatter.
- Patch 2 is:
 
    As we are doing ecosystem wide; bump the MSRV to Rust 1.48.0
    
    Includes:
    - A non-trivial clippy fix
    - Removal of the MacOS carve out in CI, no clue if this is fixed with
    Rust 1.48, we will see if ci is green.


tcharding:
Put up as part of #92 to save rebase bother.

PR #92, opened 2023-05-03
Bump MSRV and improve CI

tcharding:
Can we do all this in a single PR please because I'm too lazy to rebase #91 with fixes in the correct order. This has all the patches we need to get green CI (I think)

- run fmt
- bump MSRV
- fix clippy
- update manifest 
- bring ci into line with `rust-bitcoin`


apoelstra (approved):
ACK c35ae87e2cc1798645973f0d3201ec43974276d7

PR #93, opened 2023-05-03
Introduce SPDX license identifiers

tcharding:
Licenses are boring as hell, so is are all the comments at the top of each file. This patch makes no comment on the merit of license comments in each file, rather this patch reduces the license comment to the minimum possible with no loss of meaning - an SPDX license identifier.

Note also please that we remove the "written by" comments as well for the following reasons (discussed recently on rust-bitcoin repo):

- they are not descriptive because many devs contributed
- they have a tendency to include the wrong date because of cut'n'pasta
- all this info is in the git history

ref: https://spdx.dev/ids/#how

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,16 +1,4 @@
-// Rust JSON-RPC Library
-// Written in 2015 by
-//     Andrew Poelstra <apoelstra@wpsoftware.net>
-//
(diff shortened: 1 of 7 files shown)


apoelstra (approved):
ACK bb5b5cf5081aa5fe24effcf8bc5b562a3158b461

PR #56, opened 2022-04-25
WIP: refactor to support sync/async support

stevenroose:
This is a WIP but I publish it because it is used in a WIP refactor for the bitcoincore-rpc crate which uses this

PR #94, opened 2023-05-03
Add minreq usage

tcharding:
Add a new module `minreq_http` that implements `client::Transport` using `minreq` as the underlying HTTP transport.

- fuzzing infrastructure copied from `rust-bitcoin`
- fuzz test copied from `simple_http`
- basic integration test copied from `rust-bitcoincore-rpc`
- A public function to parse a bitcoind cookie file

https://github.com/neonmoe/minreq/


apoelstra:
Since #92 is merged maybe we can rebase and un-draft this?

apoelstra:
CI failure is because we need to add a doccomment now to the serde_json re-export. Unrelated to this PR.

Cargo.toml
```diff
@@ -19,6 +19,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 default = [ "simple_http", "simple_tcp" ]
 # A bare-minimum HTTP transport.
 simple_http = [ "base64" ]
+# A transport that uses `minreq` as the HTTP client.
+simple_minreq = [ "base64", "minreq" ]
```
apoelstra, line 23: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

This should be called `minreq_http` since `minreq` is the implementation, not the protocol.

src/simple_minreq.rs
```diff
@@ -0,0 +1,271 @@
...
+    Json(serde_json::Error),
+    /// Minreq error.
+    Minreq(minreq::Error),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        match *self {
+            Error::Json(ref e) => write!(f, "JSON error: {}", e),
+            Error::Minreq(ref e) => write!(f, "MINREQ error: {}", e),
```
apoelstra, line 131: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

Arguably the prefixes shouldn't be here at all, but if so, they should be `JSON:` and `minreq:`. There's no need for the word `error` and `MINREQ` should not be capitalized.

src/simple_minreq.rs
```diff
@@ -0,0 +1,271 @@
...
+        Builder::new()
+    }
+}
+
+impl crate::Client {
+    /// Creates a new JSON-RPC client using a bare-minimum HTTP transport.
+    pub fn simple_http_minreq(
+        url: &str,
+        user: Option<String>,
+        pass: Option<String>,
```
apoelstra, line 242: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

I think we should take `&str` rather than `String` here for both params. (Ideally it could be generic but that sucks because if the user passes `None` then the compiler will complain.)

src/simple_minreq.rs
```diff
@@ -0,0 +1,271 @@
...
+
+impl Default for Builder {
+    fn default() -> Self {
+        Builder::new()
+    }
+}
+
+impl crate::Client {
+    /// Creates a new JSON-RPC client using a bare-minimum HTTP transport.
+    pub fn simple_http_minreq(
```
apoelstra, line 239: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

Let's drop the `simple` prefix and call this `minreq_http`.

src/lib.rs
```diff
@@ -24,6 +24,9 @@ mod util;
 #[cfg(feature = "simple_http")]
 pub mod simple_http;
 
+#[cfg(feature = "simple_minreq")]
+pub mod simple_minreq;
```
apoelstra, line 28: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

Module name should be `minreq_http`.

src/simple_minreq.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// The default TCP port to use for connections.
+/// Set to 8332, the default RPC port for bitcoind.
+pub const DEFAULT_PORT: u16 = 8332;
+
+/// Simple HTTP transport that implements the necessary subset of HTTP for
+/// running a bitcoind RPC client.
+#[derive(Clone, Debug)]
+pub struct SimpleHttpTransport {
```
apoelstra, line 63: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

This struct should not be called `SimpleHttpTransport`.

src/simple_minreq.rs
```diff
@@ -0,0 +1,271 @@
...
+    url: String,
+    timeout: Duration,
+    /// The value of the `Authorization` HTTP header, i.e., a base64 encoding of 'user:password'.
+    basic_auth: Option<String>,
+}
+
+impl Default for SimpleHttpTransport {
+    fn default() -> Self {
+        SimpleHttpTransport {
+            url: format!("localhost:{}", DEFAULT_PORT),
```
apoelstra, line 73: In d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432:

minreq seems to require a `http://` prefix on this or else you get the error "was redirected to an absolute url with an invalid protocol".

apoelstra:
Done testing d20ba2eafcaa3cd5245f5a548cbcbb6d14c13432. It looks like it "just works", though it's a little bit (1-5% maybe?) slower than simple http.

OTOH if it works for Mac users without bullshit, that's a huge win.

src/simple_minreq.rs
tcharding: I"ll add integration testing against a bitcoind instance like we do in `rust-bitcoincorerpc` eh, to catch this sort of mistake.

src/simple_minreq.rs
tcharding: woops, changed to:
```rust
            Error::Json(ref e) => write!(f, "parsing JSON failed: {}", e),
            Error::Minreq(ref e) => write!(f, "minreq: {}", e),
```

src/simple_minreq.rs
tcharding: What value do these helper constructors even add? I rekon we should get rid of them from all the modules.

tcharding:
Now includes fuzzing and integration tests pulled out of `rust-bitcoincore-rpc` work by @stevenroose

tcharding:
Will need changes in https://github.com/rust-bitcoin/rust-bitcoin/pull/1842 if that merges.

apoelstra:
@tcharding I actually don't think it'll need the #1842 changes, because we still haven't changed this crate to rename `cfg(fuzzing)` to `cfg(jsonrpc_fuzzing)` and haven't updated our CI fuzzing to the new rust-bitcoin model.

tcharding:
Oh my bad, I misread the PR. The broken-crypto thing has nothing to do with fuzzing here but do we want to change this one to use `jsonrcp_fuzzing` as well just for uniformity and also in case some other reason ever turns up where we want total control of the cfg option instead of using hongfuzz's?

src/simple_minreq.rs
apoelstra: That would be amazing. TBH I usually test nontrivial PRs  to this crate by syncing my hot wallet with it, which is probably not the safest testing strategy..

src/simple_minreq.rs
apoelstra: Sounds good!

apoelstra:
@tcharding So, I did actually find myself wanting to use actual JSONRPC with network access within a fuzzer recently ... I was doing unholy things, but it did come up. So it's nice to be able to turn it off.

Maybe rather than renaming it we should add another feature gate that you have to turn *on* to re-enable the real stuff. So our `cfg(fuzzing)` guards would be `cfg(all(fuzzing, not(jsonrpc_real_fuzzing))` or something. Exact name TBD.

stevenroose:
If minreq does async, maybe you could see if it also applies to https://github.com/apoelstra/rust-jsonrpc/pull/56?

tcharding:
> cfg(all(fuzzing, not(jsonrpc_real_fuzzing))

I've opened a separate PR to tackle this idea. For now, this PR just uses `--cfg=jsonrpc_fuzz`.

tcharding:
Please don't merge without an ack from @stevenroose because I stole the integration testing from his code and also the `Auth` struct - not sure if the `Auth` struct should be included in this since it is only used in integration testing but i thought then `rust-bitcoincore-rpc` could use it from here?

apoelstra:
Can you move all the fuzz stuff to a second commit?

src/http/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// The different authentication methods for the client.
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub enum Auth {
+    /// No authentication.
+    None,
+    /// Basic authentication using (user, password).
+    UserPass(String, String),
+    /// Basic authentication using cookie file (contains 'user:password').
+    CookieFile(PathBuf),
```
stevenroose, line 32: This is really bitcoin-core specific stuff. "Basic Auth" just has a `user:pass` format and the cookiefile thing is just Bitcoind that we do in bitcoincore-rpc. I don't think it belongs here. A string field "basic_auth_header" or something should be sufficient.

stevenroose (changes requested):
I don't think we should pull in bitcoind specific stuff into this crate..

src/http/mod.rs
apoelstra: I wouldn't mind having a constructor that takes a `PathBuf` and reads a `user:pass` format out of it to construct a `UserPass`. This would let people easily use Bitcoin Core without needing the full force of bitcoincore-rpc.

But I agree we shouldn't treat the cookie file stuff as a first-class thing or directly have any data structures for it.

tcharding:
Changes in force push:
- Split PR into two patches
- Remove `Auth`, added `http::read_auth_from_file` (incl. changes to error enum)
- Cleaned up integration tests

src/http/minreq_http.rs
```diff
@@ -0,0 +1,241 @@
...
+        s.push(':');
+        if let Some(ref pass) = pass {
+            s.push_str(pass.as_ref());
+        }
+        self.tp.basic_auth = Some(format!("Basic {}", &base64::encode(s.as_bytes())));
+        self
+    }
+
+    /// Adds authentication information to the transport using a cookie string ('user:pass').
+    pub fn cookie_auth<S: AsRef<str>>(mut self, cookie: S) -> Self {
```
apoelstra, line 129: In 9b566579a2ccfd40d95056342ba45b58f14d09be:

You have this function which interprets the data from a cookie file without needing to split it then re-assemble it...

src/http/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// The Default SOCKS5 Port to use for proxy connection.
+/// Set to 9050, the default RPC port for tor.
+// Currently only used by `simple_http` module, here for consistency.
+#[cfg(feature = "proxy")]
+pub const DEFAULT_PROXY_PORT: u16 = 9050;
+
+/// Reads a 'user:password' string from `cookie`.
+///
+/// This function exists to support reading the cookie file used by `bitcoind`.
+pub fn read_auth_from_file(cookie: &Path) -> Result<(String, String), CookieError> {
```
apoelstra, line 27: In 9b566579a2ccfd40d95056342ba45b58f14d09be:

...but then you have this function which isn't compatible with the `cookie_auth` function. I think you could drop this entirely and just use `fs::read_to_string`.

fuzz/fuzz_targets/minreq_http.rs
```diff
@@ -0,0 +1,59 @@
+
+extern crate jsonrpc;
+
+#[cfg(not(jsonrpc_fuzz))]
+compile_error!("You must set RUSTFLAGS=--cfg=jsonrpc_fuzz to run these test, or run the actual fuzz harness.");
```
apoelstra, line 5: In 9710de91fd51f43aa1ac25dfd3c42e240ecb78b6:

These cause `cargo test` to fail, which will cause `cargo test --all` to fail once we move the fuzztests into a workspace.

Also, could we move the fuzztests into a workspace?

IMO we should just invert the `jsonrpc_fuzz` cfg variable and require the user *set* a cfg flag to get normal fuzzing. It's probably a bad idea to be fuzzing something with network access anyway.

tcharding: I just rebased this PR on top of #98 to make sure I got everything right. #98 is kinda trivial so you could either just merge this one and I"ll comment as such on #98 or else merge #98 if you prefer and I'll rebase this one - either way is good with me.

apoelstra:
Fuzzing stuff looks good now. Now that #98 is merged we could maybe rebase again. I think we still need to tweak the auth stuff.

tcharding:
Sure thing, thanks man.

src/simple_minreq.rs
tcharding: Added `http://` to the string, FTR I have not tested this code path.

src/simple_minreq.rs
tcharding: Can you confirm that you are ok with my removal of these helper functions mate?

tcharding: I had a quick check to make sure the new code was not non-uniform and only `simlple_http` provides these construtors on `Client` so its the odd one out not `minreq_http`.

src/http/mod.rs
tcharding: This suggestion has been implemented in the PR
```rust
/// Reads a 'user:password' string from `cookie`.
///
/// This function exists to support reading the cookie file used by `bitcoind`.
pub fn read_auth_from_file(cookie: &Path) -> Result<(String, String), CookieError> {
    ...
```

tcharding:
>  I think we still need to tweak the auth stuff.

Not quite sure what more needs doing for auth, we have the read-from-file function in `http/mod.rs` and the `minreq_http::Builder` has `pub fn auth<S: AsRef<str>>(mut self, user: S, pass: Option<S>) -> Self {`.

Note, both `minreq_http` and `simple_http` builders still have a `cookie_auth` method - is this too much?

tcharding:
Fixed up auth stuff, I ended up doing so by removing all the stuff in the `http` module and just adding an example to the `Builder::cookie_auth`  method showing how to use it. Sound ok?

Also , I pull _all_ the fuzzing stuff out of the initial patch and into the final "add fuzzing" patch.

tcharding:
Sorry bro, guess I did not run that before re-pushing. Will fix.

tcharding:
Added file reading code in the integration test and added the integration test to the workspace.

tcharding:
... and split out the integration tests to a separate patch as requested.

apoelstra:
@tcharding ok 297963bf73a759a986eb9be14e984fad7dbf6dac is almost there :)

Basically, the fuzz stuff needs to be updated -- in particular the version of honggfuzz needs to be 0.5.55 rather than 0.5, and the `honggfuzz` feature should be removed. Probably the simplest thing would be to copy the latest fuzztest infra from rust-bitcoin or rust-miniscript.

We can do this in a followup PR if you'd like, but since this one introduces fuzztesting we might as well go all the way with it.

tcharding:
Shit, super sloppy. I'll sort it in this PR. Thanks

tcharding:
Now I'm confused, what code are you looking at? In commit: `ccef3a3 Add fuzzing` I see:
```diff
-[features]
-honggfuzz_fuzz = ["honggfuzz"]
-
 [dependencies]
-honggfuzz = { version = "0.5", optional = true, default-features = false }
-jsonrpc = { path = ".." }
+honggfuzz = { version = "0.5.55", default-features = false }
+jsonrpc = { path = "..", features = ["minreq_http"] }
 ```

**Comment by apoelstra** [comment]
Yep, I'm sorry, I think I was looking at the second commit of this branch rather than the last. I think this is good to go :)

**Review by apoelstra — APPROVED** [review]
ACK ccef3a3584e5069dfbdcccb0da6825bfde32f08c

**Comment by apoelstra** [comment]
@tcharding could you open a PR to cut a release? Then I can ACK/merge it.

**Comment by tcharding** [comment]
I don't think we should blame github for that one, its not even using the correct cfg flags, my bad. Fixed in #99

**Comment by apoelstra** [comment]
I still blame github. I think if they're gonna use an opaque ad-hoc language to configure their CI, which can only be debugged via reading green checkmarks, then they shouldn't show green checkmarks on PRs that do it wrong.

Or they should use a real language so people can review the code.

File: src/http/minreq_http.rs [hunk 12]
```diff
@@ -0,0 +1,254 @@
...
+                .with_timeout(self.timeout.as_secs())
+                .with_header("Authorization", auth)
+                .with_json(&req)?,
+            None => minreq::Request::new(minreq::Method::Post, &self.url)
+                .with_timeout(self.timeout.as_secs())
+                .with_json(&req)?,
+        };
+
+        let resp = req.send()?;
+        let json = resp.json()?;
```
darosior, line 70: Shouldn't we check the `status_code` of the `resp` here?
Otherwise calling `json()` on a response that isn't `200` would try to deserialize an empty body and therefore confusingly return a `SerdeJsonError` (https://github.com/neonmoe/minreq/blob/75f28d4eba632e7652b6c15adc746a251c657ad2/src/response.rs#L164-L177) whereas it's not a JSON error but an invalid request.

apoelstra: Bitcoin Core does not return an empty body alongside HTTP error codes. It returns a JSON blob with error information.

apoelstra: I think we could just try to parse the JSON, and in the case of error, check whether the response was 200. For 200 we should return the JSON error, for non-200 we should return a status error.

Trying to match Core's exact behavior here, which I doubt is a stable part of their API, seems too fragile.

darosior: Sounds good!

PR #95, opened 2023-05-08
Add rustdoc to pub crate export

tcharding:
Fix up the extern crates by doing:
    
- Remove edition 2015 use statements
- Feature guard `base64` re-export with the correct feature
- Re-export `serde` crate
- Add rustdoc to all three pub extern crates

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,10 +11,13 @@
 
 use serde::{Deserialize, Serialize};
 
-extern crate serde;
+/// Re-export `serde` crate.
+pub extern crate serde;
+/// Re-export `serde_json` crate.
 pub extern crate serde_json;
 
-#[cfg(feature = "base64-compat")]
+/// Re-export `base64` crate.
+#[cfg(feature = "base64")]
 pub extern crate base64;
 
 pub mod client;
--- a/src/simple_http.rs
+++ b/src/simple_http.rs
@@ -15,10 +15,6 @@ use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::Duration;
 use std::{error, fmt, io, net, num};
 
-use base64;
-use serde;
-use serde_json;
-
 use crate::client::Transport;
 use crate::{Request, Response};
 
--- a/src/simple_tcp.rs
+++ b/src/simple_tcp.rs
@@ -6,9 +6,6 @@
 
 use std::{error, fmt, io, net, time};
 
-use serde;
-use serde_json;
-
 use crate::client::Transport;
 use crate::{Request, Response};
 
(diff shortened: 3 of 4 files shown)


src/lib.rs
```diff
@@ -12,6 +12,7 @@
 use serde::{Deserialize, Serialize};
 
 extern crate serde;
+/// Re-export serde_json for access to error type used in the public API of this crate.
```
apoelstra, line 15: In f3b23a09ec672d86e3cb7827872d1d338ba23239:

Let's just say `Re-export serde_json crate`. The reason given doesn't really make sense because the error type is already accessible. If you want to give a reason, you can say that it simplifies dependency management.

tcharding: I'm inclined to just put `#![allow(missing_docs)]` because writing _what_ the code does is pointless. However I'm now confused because `cargo clippy` isn't warning in my shell right now, only `cargo +nightly clippy` is and we don't run clippy with nightly in CI - so I don't know why I did this PR in the first place.

tcharding:
This warning comes from building docs not from clippy, which we do with nightly.

src/lib.rs
tcharding: So this code is silly
```rust
extern crate serde;
/// Re-export `serde_json` crate.
pub extern crate serde_json;
```

1. The comment is a _what_ comment not a _why_ comment
2. I have no idea why we pub export `serde_json` and not `serde` (from the code and just in generaly).

We could fix both of these problems by putting a comment that said why we pub export `serde_json`.

apoelstra: The why is the what in this case. We re-export the crate so that it's available. (And this lint is absolutely idiotic.)

I'm fine to just `#[allow]` the lint if you'd like.

Also we really should `pub extern crate` serde as well ... it looks like `serde_json` was exported by me in a direct push to master 4 years ago, presumably because I needed `serde_json` specifically, and then we never revisited it.

tcharding: In general any crate we depend on for types that appear in the public API, should be re-exported, right?

> (This lint is absolutely idiotic, to be clear.)

We agree on this point :)

tcharding:
Changes in force push:
- Re-write PR description
- Fix a bunch of problems with the extern crates (see PR description)

src/lib.rs
apoelstra: With some exceptions, ofc. For example `bitcoin-internals` shouldn't be re-exported.

apoelstra (approved):
ACK 1aa721550e6907239bb14d3f3840def5b36d71ab

src/lib.rs
tcharding: Cool, got it.

PR #99, opened 2023-05-27
Fix fuzzing ci job

tcharding:
Fix fuzzing introduced in #94. Patch one is cleanup, patch 2 is the fix.

Hey @apoelstra, you could consider just merging #100 since it includes this.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -37,8 +37,10 @@ simple_http,
           override: true
           profile: minimal
       - name: fuzz
-        run: if [[ "${{ matrix.fuzz_target }}" =~ ^bitcoin ]]; then export RUSTFLAGS='--cfg=hashes_fuzz --cfg=secp256k1_fuzz'; fi
-        run: cd fuzz && ./fuzz.sh "${{ matrix.fuzz_target }}"
+        run: |
+          export RUSTFLAGS='--cfg=jsonrpc_fuzz'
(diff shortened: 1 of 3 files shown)


tcharding:
CI is failing, @apoelstra:
- Why do we fuzz with Rust 1.58?
- The build fail does not happen with later versions, stable 1.69 in particular
- Any reason to not use stable so I can be lazy and not investigate why this started failing right now when the code is not new code (the mutex stuff in `simple_http` anyways)?

apoelstra:
>Why do we fuzz with Rust 1.58?

Because this is the version that works with honggfuzz. Stable stopped working (at least for a while) and earlier versions don't work.

>The build fail does not happen with later versions, stable 1.69 in particular

Yeah, we can switch to 1.69 for this crate. It looks like I'm using 1.64 locally, which also works.

> Any reason to not use stable so I can be lazy and not investigate why this started failing right now when the code is not new code (the mutex stuff in simple_http anyways)?

It started failing now because this is the first time we're running the fuzz tests in CI, and the offending code is gated on the fuzz cfg flag. I encountered this a few weeks ago after merging #98, and forgot about it, sorry.

But no, let's not use stable, that'll break. Let's just pick a version that works.

tcharding:
All green - party.

apoelstra (approved):
ACK 87338f3677364ca377a887c53673fcdb2513a445

apoelstra:
Will just review/merge #100.

PR #102, opened 2023-06-28
minreq_http: re-export minreq dependency

darosior:
So that downstream users can match against error variants.

I had to add minreq as a direct dependency to my project. Since we are using `minreq::Error` in one variant of `minreq_http::Error`, better let users directly access it as `minreq_http::minreq::Error`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,10 @@ pub extern crate serde_json;
 #[cfg(feature = "base64")]
 pub extern crate base64;
 
+/// Re-export `minreq` crate if the feature is set.
+#[cfg(feature = "minreq")]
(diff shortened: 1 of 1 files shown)


apoelstra:
I think it'd be a bit more idiomatic to add a
```
#[cfg(feature = "minreq")]
pub extern crate minreq;
```
to src/lib.rs. You can see how this is already done for base64.

darosior:
Done.

apoelstra:
Can you do `pub extern crate` rather than `pub use` so it's obvious that this is a crate re-export?

apoelstra (approved):
ACK 6a5a578338c7779a1ebec98c37f31ca8845faa8f

PR #103, opened 2023-06-28
minreq_http: return an HTTP error on error with no JSON in body

darosior:
It is useful for downstream users to be matching on errors that do not contain a valid JSONRPC error in the HTTP response body. One instance is if the HTTP server work queue depth is exceeded, as they probably want to retry the request later.

On such error we would return a JSON deserialization error, without exposing neither the HTTP status code nor the body of the response. This made it impossible to detect such transient errors.

Instead, introduce a new HttpError variant that gets returned when the requested is responded to by an error that does not contain valid JSON in its body.

See also: https://github.com/apoelstra/rust-jsonrpc/pull/94#discussion_r1245103898.

### DIFF
--- a/src/http/minreq_http.rs
+++ b/src/http/minreq_http.rs
@@ -66,9 +66,23 @@ impl MinreqHttpTransport {
                 .with_json(&req)?,
         };
 
+        // Send the request and parse the response. If the response is an error that does not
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 66fb441ca2d6d1a31ac07a4c374e046d868adaee

PR #108, opened 2023-11-09
Accept also None or Objects as params

### LINKED ISSUE #107: Support params as Object
This lib is accepting only arrays as params:

https://github.com/apoelstra/rust-jsonrpc/blob/b8c3f0d4d78440a286dd1fa7e97d6f78520b16b6/src/lib.rs#L73

I think because that was the only way available to communicate with core.
However, now there are named parameters and to support those this library should supportalso json objects as params.

Indeed jsonrpc spec https://www.jsonrpc.org/specification#request_object allows it to be an array or an object

RCasatta:
according to the jsonrpc spec params can be missing, or an object too.

close #107 

Breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.17.0 - 2023-06-29
+
+* `params` field in `Request` changed to a generic `RawValue` instead of an array.
+  [#108](https://github.com/apoelstra/rust-jsonrpc/pull/108)
+
 # 0.16.0 - 2023-06-29
 
 * Re-export the `minreq` crate when the feature is set
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ fn client(url: &str, user: &str, pass: &str) -> Result<Client, simple_http::Erro
 // Demonstrate an example JSON-RCP call against bitcoind.
 fn main() {
     let client = client("localhost:18443", "user", "pass").expect("failed to create client");
-    let request = client.build_request("uptime", &[]);
+    let request = client.build_request("uptime", None);
     let response = client.send_request(request).expect("send_request failed");
 
     // For other commands this would be a struct matching the returned json.
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -16,7 +16,11 @@ fi
 # Pin dependencies as required if we are using MSRV toolchain.
 if cargo --version | grep "1\.48"; then
     # 1.0.157 uses syn 2.0 which requires edition 2021
+    cargo update -p serde_json --precise 1.0.99
     cargo update -p serde --precise 1.0.156
+    cargo update -p quote --precise 1.0.30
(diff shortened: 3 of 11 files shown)


apoelstra:
concept ACK. But looks like certain feature combos are broken (see CI)

RCasatta:
I thought after the first approval the other times the CI would start automatically, but it's not the case, I hope to have "blindly" fixed it.

RCasatta:
I think to have tested all the cases launched with the CI locally, sorry for the back and forth

apoelstra:
>I thought after the first approval the other times the CI would start automatically, but it's not the case, I hope to have "blindly" fixed it.

All good. I blame Github, not you :).

Hit CI. Let's see how it goes.

apoelstra (approved):
ACK 2c3f4622a5fe7df7b434d163267d4cb21d246cdd

apoelstra:
@RCasatta can you add a unit test that demonstrates how to provide a list, and maybe a CHANGELOG entry explaining this? This will be a breaking change which will be particularly annoying because the `RawValue` API is so annoying.

RCasatta:
I had to force push because there were a couple of reference of `&[]` not changed and not tested in CI (the one in the Readme and the one in `integration_test/src/main.rs`)

Moreover, I added:
* the unit test commit
* a call `getblockhash` in the integration test with both list and named configuration (not launched in CI though)
* update the changelog

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.17.0 - 2023-06-29
```
tcharding, line 1: nit: Perhaps use "unreleased" so we don't forget to update the header when doing the release.

```suggestion
# unreleased
```

tcharding (approved):
ACK  178ee93f487157da3efec5f9632ca8fb5de6d9f9

apoelstra:
Looks like some clippy lints are causing CI to fail.

RCasatta:
Sorry again, should be fixed

tcharding (approved):
ACK be69ad162d76fdecfb0e3b940aae18f031ff3a83

apoelstra:
Previously you had to provide a list with each thing wrapped in `arg()`, which was bad enough ... but now it looks like you need to wrap your lists in `Some(to_raw_value(&json!(actual_thing)).unwrap())` which is way worse.

I think we need a better API here.

RCasatta:
You don't need the `json!` call when you have a `struct` with `serde::Serialize`

apoelstra:
@RCasatta ah, I gotcha. So in your `list` unit test, where you write `json!([0])` it's actually fine to just do `to_raw_value(&[0])` (I tested this and it worked).

Ok, I'm happy then.

apoelstra (approved):
ACK 2c3f4622a5fe7df7b434d163267d4cb21d246cdd

PR #114, opened 2024-02-05
SimpleHttpTransport error on chunked transfer encoding

### LINKED ISSUE #112: simple http error when transfer encoding is chunked
When trying to decode a big json from a json rpc server returns the fllowing error:

```
    0: JSON decode error: invalid type: integer `2000`, expected struct Response at line 1 column 5
```

while `curl` opens the url correctly and the body parse to a json.

By looking at headers with `curl` I noted the transfer encoding is chunked and that may be the cause of the issue

RCasatta:
close #112


apoelstra (approved):
ACK 575c8b4057c8a055bca67f4fbcb1f7fc06095991 Thanks for doing this!

PR #118, opened 2024-05-17
Allow assigning_clones

tcharding:
`clone_into` is only available in Rust 1.63 and since our MSRV is Rust 1.48 just shoosh clippy

### DIFF
--- a/src/http/minreq_http.rs
+++ b/src/http/minreq_http.rs
@@ -121,6 +121,7 @@ impl Builder {
     }
 
     /// Sets the URL of the server to the transport.
+    #[allow(clippy::assigning_clones)] // clone_into is only available in Rust 1.63
     pub fn url(mut self, url: &str) -> Result<Self, Error> {
         self.tp.url = url.to_owned();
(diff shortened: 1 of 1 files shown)


tcharding:
Can we merge this with red CI? Its the cfg thing, needs nightly pin which is done in #119 but #119 needs this one.

apoelstra:
> `clone_into` is only available in Rust 1.63 and since our MSRV is Rust 1.48 just shoosh clippy

Is there a bug about `clone_into` lint not respecting the MSRV in clippy.toml?

apoelstra:
There is, which was fixed 2 months ago: https://github.com/rust-lang/rust-clippy/issues/12502

The lint has also been moved to "pedantic" and allow-by-default. https://rust-lang.github.io/rust-clippy/master/index.html#/assigning_clones

I'm confused as to why it's affecting us now.

apoelstra:
Oh, I see, this made it into stable :(.

tcharding:
Now in #121

PR #117, opened 2024-05-05
Delete `.rustfmt.toml`

tcharding:
In commit `14429b13` I introduced a new config file for rustfmt but did not delete the hidden `.rustfmt.toml` so the new file had no effect.

- Delete the old confid `.rustfmt.toml`
- Make `rustfmt.toml` mirror the current one in `rust-bitcoin` with the exception of `edition = 2018`
- Run `cargo +nightly fmt`

Fix: #116

PR #119, opened 2024-05-17
Use rust-bitcoin-maintainer-tools and re-write CI

tcharding:
As we have been doing in other crates, use the new maintainer tools test script we created from `rust-bitcoin/rust-bitcoin-maintainer-tools/ci`.

- This should not change test coverage in any meaningful way.
- The integration tests are not currently run in CI, this patch preserves that behaviour.
- Introduces recent/minimal lock files.


apoelstra:
Looks like the `serde_json` version is too low and some other crates' versions are too high, in the lockfiles.

We also probably want to pin nightly rustc in the same PR.

tcharding:
Note please the recent and minimal lock files are the same. I spent 10 minutes trying to fix them otherwise, but my fingers started to bleed.

Also, I'll fix the lint job after #117 goes in.

tcharding:
I don't see whats up with the lint job, I can't reproduce locally?

apoelstra:
@tcharding it looks like your lockfiles have `serde` 1.0.101, which does not compile anymore.

It looks like serde 1.0.103 will work, which is what we have in the Cargo.toml for `bitcoin` and `units` and `internals` (`hashes` has 1.0 which is a bug, oops). I remember the minimal version needing to be even higher...our Cargo-minimal.lock in the rust-bitcoin repo is at 1.0.156, which is suspicious because the minimal lockfile should match Cargo.toml.

Anyway, to get this PR compiling we just need to bump the serde version. But separately I should go audit the lockfiles in rust-secp and rust-bitcoin.

tcharding:
Since this crate is MSRV 1.48 still, I wasn't able to use the other repos lock files for guidance. Simply using `cargo update -p serde --precise 1.0.103` doesn't seem to work here. I'm on an epic fast to try get over all the illnesses I keep getting clobbered with at the moment so I'll come back to this when I'm eating again.

tcharding:
Oh the MSRV thing is going to be a problem if we want to move this to `bitcoind-json-rpc` repo, any reason not to bump this crate to 1.56.1?

apoelstra:
Nope, no reason. Let's do it.

tcharding:
Now includes the MSRV bump, are you ok to do that in this PR or is that too much going on?

tcharding:
Not sure whats up with the lint fail.

apoelstra:
The lint failure looks like an outdated version of serde is being used.

tcharding:
oh thanks, nice. I'll try to fix that up.

apoelstra:
Still having trouble with your lockfiles. I will try to figure it out and submit a diff.

integration_test/Cargo.toml
```diff
@@ -8,5 +8,5 @@ edition = "2018"
 jsonrpc = { path = "..", features = ["minreq_http"] }
 lazy_static = "1.4.0"
 log = "0.4"
-backtrace = "0.3.50"
+backtrace = "0.2.3"
```
apoelstra, line 11: In 10972a7846412d323911b382010a657a539b5296

Why did you change this? backtrace 0.2 does not link for me.

tcharding: No idea what I was doing, now removed. Note the recent and minimal files are the same.

apoelstra:
OK, this is looking better. The new lockfiles still don't work with 1.48.0 but at least this is consistent when running locally `cargo +1.48.0 test` and in CI, rather than being a mysterious nix-only thing.

You will probably have a better time with this if you first bump the MSRV to 1.56.1.

tcharding:
Oh I did not realize we were on 1.48 here still. I'll bump to 1.63

tcharding:
Rebased to pick up the MSRV bump. Also added a patch upgrading `honggfuzz`.

integration_test/Cargo.toml
apoelstra: In a065c9fe4b009dc2079cbb2970a83213b602b1cb:

This is not removed.

contrib/crates.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# Crates in this workspace to test (excl. fuzz an integration-tests).
+CRATES=(".")                    # Non-workspaces don't have crates.
```
apoelstra, line 4: In a065c9fe4b009dc2079cbb2970a83213b602b1cb:

What is going on with this file? What does "non-workspaces don't have crates" mean, and does this actually exclude testing the `fuzz` and `integration-test` crates, and why?

tcharding: Oh yes you are correct this should include `fuzz`, and I have that wrong in `miniscript` as well. The comment is also wrong and was always wrong, I wrote "Non-workspcase" when I meant "repos that have crate root in the same directory as repo root" (I don't know the name for that). We need to keep the "." so it should be 
```
CRATES=("." "fuzz")     # Some comment about the dot directory.
```

apoelstra: I think you want then

```
# Directories of each workspace to test
CRATES=("." "fuzz")
```

Saying "crates in this workspace" is just confusing. Workspaces are crates.

apoelstra: BTW the reason I'm complaining about this is that the existing lockfiles don't work with MSRV, which you can tell if you try to compile the `integration_test` workspace.

tcharding: I think I also forgot that this crate is 1.48 still, at least that was not in my mind yesterday.

tcharding: oh there's the problem, I bumped the MSRV in this PR - I'll pull it out into a separate one.

tcharding: Alright, I'm well confused. I copied `Cargo-recent.lock` to `Cargo.lock` and cd'ed into `integration_test` and ran
```
 cargo +1.63.0 --locked check
```

And it was ok.

tcharding: I just went right ahead and added `fuzz` and `integration_test`  to `CRATES` also.

Note that we do not run the integration tests now or previously.

tcharding: > I think I also forgot that this crate is 1.48 still

What the hell was I thinking

integration_test/Cargo.toml
apoelstra: In d4ed55570290e378aee957fdf006d0ce0bb1749b:

Please remove this.

tcharding: My bad. Done, again.

apoelstra:
If you run `cargo +1.63.0 test --all-features --locked` in the integration test directory, none of `object` `backtrace` `addr2line` (at least) compile. In 04dc4b2ffe6af33feb5481e65bc03f3b9b09ae9a.

tcharding:
hmm, something is not quite right then because that should have been caught by CI. I'll investigate further.

tcharding:
Found the bug in `run_task`, we have a `break` instead of a `continue` in a loop. Funny after we just hit that same mistake in the bech32 the other day. We were not hitting it before because `fuzz` was always at the end of the list of `CRATES`.

apoelstra:
Oh, neat. I feel much better about what a PITA this PR has been. Apparently we needed to use the maintainer tools on a crate whose CI was a total mess :).

tcharding:
Ok, this one should be right to merge if it gets past your local CI.

apoelstra (approved):
ACK 8f1779447131de57e0f32797bfa16209694d1208 successfully ran local tests

PR #121, opened 2024-05-17
integration_test: Clear clippy warnings

tcharding:
Clear all the `clippy` warnings from the `integration_test` crate. Done in preparation for re-writing CI to use the new `run_task` script. The last patch is bodgy.

### DIFF
--- a/integration_test/src/main.rs
+++ b/integration_test/src/main.rs
@@ -143,7 +143,7 @@ fn main() {
 
 fn test_get_network_info(cl: &Client) {
     let request = Request {
-        method: "getnetworkinfo".into(),
+        method: "getnetworkinfo",
         params: None,
         id: serde_json::json!(1),
         jsonrpc: Some("2.0"),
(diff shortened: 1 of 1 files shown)


apoelstra:
Last patch is interesting. The code in question is unsound. May be worth filing a compiler bug showing that the `take` code triggers the warning but the `replace` code does not, because I believe they are exactly equivalent.

Meanwhile, we should replace this `unsafe { &mut RESULTS }` shit with a mutex. Will need to use a `OnceCell` or something to initialize it.

apoelstra:
Looks like bitcoincore-rpc, where we copied this from, now just uses `lazy_static`. We might as well copy the same solution. **Or**, we can use `Mutex::new()` in a const context, but only since 1.63, so if we did this, we'd no longer be able to run the integration tests with MSRV.

apoelstra:
I propose:

* We PR to fix the unsoundness (and nothing else)
* Then we rebase this PR, and add a commit which whitelists the broken `assigning_clones` lint.
* Then we rebase the "rewrite CI" PR, adding whatever we need to pin the nightly version.

Then I think we should get green CI on all three.

tcharding:
Thanks for putting some time into this pile of PRs - I'll have a ago at your suggestion week (its the weekend now for me).

tcharding:
Bump please.

apoelstra (approved):
ACK ac3fa84c6b8fe62a5ce825e0a107a2cd2a1710bc

PR #125, opened 2024-08-01
Bump MSRV to 1.63.0

tcharding:
As we did in `rust-bitcoin` and because 1.63.0 is in Debian stable; bump the MSRV to Rust `v1.63.0`.

Remove the docs about pinning but leave the CI script unchanged because we are going to remove it shortly anyways.


apoelstra (approved):
ACK 50c580262dcd6f6ec77f58f2529a6be2d45b1b69

PR #128, opened 2024-09-04
Add https support to minreq_http

spartucus:
This way, `bitcoincore-rpc` can use `jsonrpc`  with feature `minreq_http` for https connection.


apoelstra:
I guess I forgot to archive this repo. It now lives in rust-bitcoind-json-rpc: https://github.com/rust-bitcoin/rust-bitcoind-json-rpc/pull/19

